Issue #68, opened 2026-03-04
feat(qol): Kubectl Plugin for NodeDiagnostic Logs!


cdirubbio:
**What would you like to be added**: A kubectl plugin that allows for easy log collection

**Why is this needed**: The log collection process can be difficult now and so I'm making a PR for a kubectl plugin that customers can utilize to ease the process.

**Are you currently working around this issue?**:

How are you currently solving this problem?

**Additional context**:

Anything else we should know?

PR #69, opened 2026-03-04
feat: kubectl ekslogs plugin for NodeDiagnostic Log Collection

cdirubbio:
**Issue #, if available**: https://github.com/aws/eks-node-monitoring-agent/issues/68

**Description of changes**: When troubleshooting EKS issues, it is very often useful to acquire and download the node logs, especially when working with AWS Support Engineering.

The goal of this addition is to simplify the log collection process by providing customers with a kubectl plugin that allows for immediate collection without a set of steps to follow. I've used a non-cleaned-up version of this plugin for a while on my clusters and it's a big help

### Examples (from README):

**Collect logs from a single node:**
```bash
kubectl ekslogs ip-10-0-1-42.ec2.internal
```
**Collect logs from multiple nodes:**
```bash
kubectl ekslogs ip-10-0-1-42.ec2.internal ip-10-0-1-43.ec2.internal
```
**Collect logs from all nodes in a node group:**
```bash
kubectl ekslogs -l eks.amazonaws.com/nodegroup=my-node-group
```
**Upload logs to S3 bucket (root directory):**
```bash
kubectl ekslogs --s3 my-bucket ip-10-0-1-42.ec2.internal
```
**Upload logs to S3 with custom key prefix:**
```bash
kubectl ekslogs --s3 my-bucket --key logs/2024-03 ip-10-0-1-42.ec2.internal
```


tools/kubectl-ekslogs/kubectl-ekslogs
```diff
@@ -0,0 +1,476 @@
...
+    if ! kubectl get node "$node" >/dev/null 2>&1; then
+        error "Node not found: ${node}"
+        exit 1
+    fi
+done
+info "All ${#NODE_NAMES[@]} node(s) validated"
+
+# --- Main ----------------------------------------------------------------
+
+trap cleanup INT TERM
```
prasad0896: Should this be `trap cleanup EXIT INT TERM` incase the script fails to run?

tools/kubectl-ekslogs/kubectl-ekslogs
```diff
@@ -0,0 +1,476 @@
...
+    destination: node
+EOF
+    fi
+    ((node_index++))
+done
+
+# Wait for completion
+status "Waiting for log collection to complete (timeout: ${TIMEOUT})..."
+for node in "${NODE_NAMES[@]}"; do
+    if ! kubectl wait \
```
prasad0896: This seems fragile to only wait for the diagnostic cr status. The status will be set regardless of success or failure. We should wait for success status as well in here.

tools/kubectl-ekslogs/kubectl-ekslogs
cdirubbio: agreed yep, fixed in new commit

PR #82, opened 2026-03-10
chore: switch dcgm-exporter to distroless image

jsonmp-k8:
## Summary
- Switches dcgm-exporter container image from `ubuntu22.04` to `distroless` variant (`4.5.2-4.8.1-distroless`)
- Reduces attack surface by removing shells, package managers, and OS utilities from the container
- Updates container command from shell-based (`/bin/sh -c "nv-hostengine ..."`) to direct binary execution (`/usr/bin/nv-hostengine -n -b ALL`), since distroless images have no shell
- Removes accidentally committed `README.md-e` backup file

## Changes
| File | Change |
|------|--------|
| `charts/.../values.yaml` | Tag: `ubuntu22.04` -> `distroless` |
| `charts/.../templates/dcgm-daemonset.yaml` | Direct binary exec instead of `/bin/sh` |
| `charts/.../README.md` | Update default tag in docs |
| `e2e/setup/manifests/agent.tpl.yaml` | Update hardcoded image tag and command |

## Why distroless
- Smaller image size (no OS package layer)
- Reduced CVE exposure (no shell, no apt/dnf, no coreutils)
- Same DCGM version (4.5.2) and CUDA version (4.8.1) — only the base layer changes
- `nv-hostengine -n` runs in foreground mode natively, so no `sleep infinity` fallback is needed

## Verification
- **Binary path confirmed**: `/usr/bin/nv-hostengine` is the correct path — it is the upstream NVIDIA DCGM Dockerfile's [ENTRYPOINT](https://github.com/NVIDIA/DCGM/blob/master/docker/Dockerfile.ubuntu22.04), consistent across both Ubuntu and distroless variants
- **Removed `|| true; sleep infinity`**: The old command silently swallowed `nv-hostengine` failures and kept the container alive via `sleep infinity`. With direct execution, if `nv-hostengine` exits on error, Kubernetes will restart the container (CrashLoopBackOff), which is more visible and correct behavior. The DaemonSet's node affinity restricts scheduling to GPU instance types only (g3/g4/g5/g6/p2/p3/p4/p5/p6 families), so non-GPU node edge cases do not apply.

## Test plan
- [ ] Deploy dcgm-server DaemonSet on a GPU node and verify `nv-hostengine` starts on port 5555
- [ ] Confirm the monitoring agent connects to DCGM and collects GPU metrics
- [ ] E2E tests pass with updated manifest

### DIFF
--- a/charts/eks-node-monitoring-agent/README.md
+++ b/charts/eks-node-monitoring-agent/README.md
@@ -37,13 +37,11 @@ The following table lists the configurable parameters for this chart and their d
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | dcgmAgent.affinity | object | see [`values.yaml`](./values.yaml) | Map of dcgm pod affinities |
-| dcgmAgent.image.account | string | `"602401143452"` | ECR repository account number for the dcgm-exporter |
-| dcgmAgent.image.containerRegistry | string | `""` | Full container registry URL override (e.g., 602401143452.dkr.ecr.us-west-2.amazonaws.com). When set, this takes precedence over account/endpoint/region/domain fields. |
-| dcgmAgent.image.domain | string | `"amazonaws.com"` | ECR repository domain for the dcgm-exporter |
-| dcgmAgent.image.endpoint | string | `"ecr"` | ECR repository endpoint for the dcgm-exporter |
+| dcgmAgent.image.override | string | `""` | Full image override (registry/repository:tag). Takes precedence over all other fields. |
 | dcgmAgent.image.pullPolicy | string | `"IfNotPresent"` | Container pull policy for the dcgm-exporter |
-| dcgmAgent.image.region | string | `"us-west-2"` | ECR repository region for the dcgm-exporter |
-| dcgmAgent.image.tag | string | `"4.5.2-4.8.1-ubuntu22.04"` | Image tag for the dcgm-exporter |
+| dcgmAgent.image.registry | string | `"nvcr.io/nvidia/k8s"` | Container image registry for the dcgm-exporter (default: NVIDIA public catalog) |
+| dcgmAgent.image.repository | string | `"dcgm-exporter"` | Container image repository name for the dcgm-exporter |
+| dcgmAgent.image.tag | string | `"4.5.2-4.8.1-distroless"` | Image tag for the dcgm-exporter |
 | dcgmAgent.podAnnotations | object | `{}` | Pod annotations applied to the dcgm exporter |
 | dcgmAgent.podLabels | object | `{}` | Pod labels applied to the dcgm exporter |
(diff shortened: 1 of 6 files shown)


prasad0896 (changes requested):
Thanks for the PR, @jsonmp-k8. The move to distroless is a good suggestion. A few things need to be addressed before this can move forward:

The distroless tag (4.5.2-4.8.1-distroless) doesn't exist in our ECR repository (
602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/observability/dcgm-exporter) that the chart and E2E manifests reference. This is the root cause of the CI
failures — the container image simply can't be pulled.

In order to fix that, I suggest taking one of the two paths forward:

1. Make the image source configurable as a build/deploy arg — allow the full image (registry + tag) to be overridden so that CI and users can point to a
registry that hosts the distroless variant. This keeps the ECR default but unblocks testing and custom deployments.

2. Pull from NVIDIA's public catalog directly — the distroless variant is available at nvcr.io/nvidia/k8s/dcgm-exporter:4.5.2-4.8.1-distroless. The chart
and E2E manifests could reference this upstream source, or we could mirror it into our ECR first and then land this PR.

Either way, the image needs to actually be pullable before CI can pass.

Helm template not updated

charts/eks-node-monitoring-agent/templates/dcgm-daemonset.yaml still has:
yaml
```
command: ["/bin/sh"]
args: ["-c", "nv-hostengine -n -b ALL || true; sleep infinity"]
```
Distroless images have no /bin/sh. This needs to change to direct binary execution:
yaml
```
command: ["/usr/bin/nv-hostengine"]
args: ["-n", "-b", "ALL"]
```
The same applies to e2e/setup/manifests/agent.tpl.yaml — the 4th commit restored the shell command but kept the distroless tag, which is contradictory.

E2E fallback is too broad

The updated dcgmiExec catches all pod exec errors and silently falls back to host exec. This will mask real failures (RBAC issues, pod not ready, dcgmi
command errors). Consider narrowing the fallback to only trigger on "executable file not found" / "no such file or directory" errors, and let other
failures propagate.

HusainZafar:
Hi @jsonmp-k8, thanks for the PR and for iterating on this!

Looks like the distroless image is now available in our ECR, which I confirmed by running:
```
docker manifest inspect \
    602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/observability/dcgm-exporter:4.6.0-4.8.3-distroless
```

Could you please update the PR to rely on this image (4.6.0-4.8.3-distroless from our ECR), remove the nvcr.io / NVIDIA public catalog logic added in the latest commit, and rebase against main to get this back to a good state?

Do you think you'd be able to pick this up soon?

prasad0896:
We merged #210 and that should address the ask in this PR as we no longer have a dependency on dcgm-exporter.

PR #88, opened 2026-03-12
Switch runtime image to Chainguard to eliminate container CVEs

jsonmp-k8:
## Summary

- Replace `eks-distro-minimal-base-glibc:latest-al23` runtime image with `cgr.dev/chainguard/glibc-dynamic:latest` to eliminate known CVEs in the final container image
- Switch builder stages from `amazonlinux:2023` to `amazonlinux:2023-minimal` to reduce builder image size (~70MB vs ~175MB)
- Upgrade Go from 1.25.5 to 1.25.8 to pick up security patches (html/template, net/url, os)
- Adjust shared library copy paths from `/usr/lib64/` to `/usr/lib/` to match Chainguard/Wolfi filesystem layout
- Add `USER 0` to override Chainguard's nonroot default (agent requires root for chroot, host filesystem access, dbus)

## Motivation

The current `eks-distro-minimal-base-glibc:latest-al23` runtime image inherits AL23 system-level vulnerabilities, including glibc CVEs (e.g. CVE-2026-0915 — stack content leak in `getnetbyaddr`, CVE-2026-0861 — heap corruption in memalign). While the image is already minimal, these CVEs are present in the base layer and cannot be mitigated without switching base images.

[Chainguard's `glibc-dynamic`](https://images.chainguard.dev/directory/image/glibc-dynamic/overview) is purpose-built for this use case:

| | eks-distro-minimal-base-glibc | chainguard/glibc-dynamic |
|---|---|---|
| Known CVEs | Inherits AL23 glibc/system CVEs | Zero or near-zero (rebuilt daily) |
| Shell / package manager | None | None |
| glibc included | Yes | Yes |
| Multi-arch (amd64/arm64) | Yes | Yes |
| Free to use | Yes | Yes (public catalog) |

## Compatibility

- **glibc ABI**: glibc maintains strict backward ABI compatibility. The systemd and DCGM `.so` files copied from the AL2023 builder stages work correctly on Chainguard's newer glibc.
- **Builder stages slimmed**: Both AL2023 builder stages (systemd-builder, dcgm-builder) switched from `amazonlinux:2023` to `amazonlinux:2023-minimal`. The minimal image includes `dnf`, `sed`, and `coreutils` — all dependencies used during the build. These stages only extract `.so` files and do not affect the final image's vulnerability surface.
- **Library paths**: Chainguard/Wolfi uses `/usr/lib/` as the standard dynamic linker search path instead of AL23's `/usr/lib64/`. All `COPY` destinations updated accordingly.
- **Root user**: Chainguard images default to nonroot (UID 65532). Added explicit `USER 0` since the agent requires root for chroot operations, host filesystem access, and dbus socket communication. This matches the previous image's default behavior.
- **Go builder**: No slim variant available — Alpine uses musl libc which is incompatible with CGO + `libsystemd-dev` (glibc). Left as-is.

## Test plan

- [ ] Build the Docker image locally: `docker build -t eks-node-monitoring-agent .`
- [ ] Verify the binary starts: `docker run --rm eks-node-monitoring-agent --help`
- [ ] Verify shared library resolution with `LD_DEBUG=libs`
- [ ] Run e2e tests on an EKS cluster with GPU and non-GPU nodes
- [ ] Scan the new image with a vulnerability scanner (e.g. `trivy image`, `grype`, or Amazon Inspector) and confirm reduced CVE count

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@
 # =============================================================================
 # Stage 1: Amazon Linux builder for systemd libraries
 # =============================================================================
-FROM public.ecr.aws/amazonlinux/amazonlinux:2023 AS systemd-builder
+FROM public.ecr.aws/amazonlinux/amazonlinux:2023-minimal AS systemd-builder
 
 RUN dnf install -y systemd-devel && \
     dnf clean all
@@ -14,7 +14,7 @@ RUN dnf install -y systemd-devel && \
 # =============================================================================
 # Stage 2: DCGM builder for GPU monitoring libraries
 # =============================================================================
-FROM public.ecr.aws/amazonlinux/amazonlinux:2023 AS dcgm-builder
+FROM public.ecr.aws/amazonlinux/amazonlinux:2023-minimal AS dcgm-builder
 
 # Install DCGM from NVIDIA repository for GPU monitoring support
 # This is optional - the agent works without it on non-GPU nodes
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,7 +83,8 @@ COPY --from=go-builder /workspace/bin/chroot /opt/bin/chroot
 # Set working directory
 WORKDIR /opt/bin
 
-# Run as non-root user (the agent will use privileged container settings for host access)
-# Note: Some operations require privileged mode, configured via Helm chart securityContext
+# The agent requires root for chroot, host filesystem access, and dbus operations.
+# Chainguard images default to nonroot (UID 65532); override for this privileged agent.
+USER 0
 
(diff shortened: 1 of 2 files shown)


prasad0896 (changes requested):
As mentioned by Micah previously, we can't merge this PR as we can't use 3rd party images.

prasad0896:
As mentioned by @micahhausler above, we cannot switch the runtime image so I am going to close this PR. Please feel free to open a new PR if you want to parameterize the arguments to pass in your own image at build time.

PR #89, opened 2026-03-13
feat: allow users to allowlist custom iptables rules

dcoppa:
**Issue #, if available**:

**Description of changes**:

- Add allowedIPTablesChains configuration option to the networking monitor to suppress false-positive UnexpectedRejectRule warnings triggered by REJECT/DROP rules in known-good custom iptables chains.
- This prevents false positive `UnexpectedRejectRule` events for legitimate custom rules.
- Updated documentation in `node-health-issues.adoc` to describe the allowlisting mechanism.

Why: we have a custom daemonset that inserts some iptables rules on the cluster nodes to block IPv4 traffic to certain subnets and we don't want eks-node-monitoring-agent to generate "NetworkingReady" events because of these custom rules.

**Testing Done**:

- Unit tests for IsExpectedRejectRule with and without custom chain names (parser_test.go)
- Unit tests for GetAllowedIPTablesChains covering empty map, missing networking entry, and populated chains (monitor_test.go)
- Validation tests: empty string for chain names and chains on non-networking monitor both return errors


dcoppa:
> Thanks for the contribution @dcoppa and having custom daemonsets that insert intentional DROP/REJECT rules and getting false positive `UnexpectedRejectRule` events is not ideal.
> 
> That said, I have concerns about using iptables comments as the allowlisting mechanism. The core issue is that this moves the trust boundary into user configuration i.e. any process on the node with CAP_NET_ADMIN (or root) can insert an iptables rule with the magic comment string and silently bypass NMA's detection. That's exactly the class of thing this check is designed to catch.
> 
> A few specific concerns:
> 
>     * The check uses strings.Contains(rule.comment, "allowed-in-eks-node-monitoring-agent"), so any comment containing that substring passes - including accidental matches. There's no structural validation of what's being allowlisted.
> 
>     * The magic string is publicly documented, so any attacker and in effect their LLM, who reads the NMA docs knows exactly what comment to add to evade detection.
> 
>     * There's no audit trail when a rule is suppressed this way - no log entry, no event, no metric. You lose visibility into what's being silently ignored.
> 
>     * Any compromise that gets root on the node (container escape, rogue daemonset, etc.) can insert harmful DROP rules and tag them to avoid detection - which defeats the purpose of the monitor.
> 
> 
> We'd prefer to solve this through configuration rather than iptables comment conventions. The agent already has per-monitor config via Helm values (nodeAgent.monitors.networking). We could extend this to accept an allowlist of reject rules to ignore, keyed on structural properties like chain name, source/destination CIDR, or protocol

Good morning,

This new implementation should address all the points and concerns you raised.

A question: is it possible to have this new configuration exposed in the add-on schema? I'd like to be able to install eks-node-monitoring-agent as an EKS add-on.

prasad0896:
>A question: is it possible to have this new configuration exposed in the add-on schema? I'd like to be able to install eks-node-monitoring-agent as an EKS add-on.

Yes, you already covered it after you made changes to  `charts/configuration.schema.json` . You can use this to override the values when installing as EKS addon.

monitors/networking/monitor.go
```diff
@@ -102,6 +103,16 @@ func WithRuntimeContext(runtimeContext *config.RuntimeContext) Option {
 	}
 }
 
+func WithAllowedIPTablesChains(chains []string) Option {
```
tzneal: Not used anywhere that I can see

dcoppa: Removed, thank you

Issue #103, opened 2026-03-19
feat(docs): add NodeDiagnostics logs documentation


youwalther65:
**What would you like to be added**:
I suggest adding NodeDiagnostic logs documentation

**Why is this needed**:
To get familiar with what NodeDiagnostics collects and use it for investigating EKS worker nod eissues.

**Are you currently working around this issue?**:

How are you currently solving this problem?

Looking in the code and the tar.gz file itself.

**Additional context**:

Anything else we should know?
This documentation could  be used as a self-service or for AWS support persona.

PR #104, opened 2026-03-19
Added NodeDiagnostics logs documentation

youwalther65:
**Issue** #103 

**Description of changes**:
Added comprehensive docs/NodeDiagnostics section with code and man pages references, sample output and detailed explanation what and how it was collected

**Testing Done**:
N/A


prasad0896 (commented):
@youwalther65 thanks for taking this on -- there's no map of what lands in a NodeDiagnostics bundle today, so #103 is worth closing.

My concern is upkeep. A hand-written README per folder duplicates what's already in the collectors, and it has to be re-edited every time one changes. It's already behind: `gpu/` (Nvidia) and `automode/` (AutoMode) ship today and write their own top-level folders, but neither is in the index, and the `var_log/` table is missing `fsx-openzfs-csi-*` and `file-cache-csi-*`. If it's drifted before merging, it'll keep drifting.

sosreport handles this by making the bundle describe itself: each file is named after the command that produced it, and the run writes a `manifest.json` listing every file, its command, and whether it ran. It can't go stale because the collection code emits it. You already have the inputs -- `collectorMap` in `pkg/logcollection/collectors.go` knows each file, its command, and the tags that gate it. Generate the index from that, fail CI when it's out of date, and a new collector documents itself.

Then hand-write only what a manifest can't explain: IPAMD's introspection JSON, the throttling files computed from cgroup stats, logdog. For `dmesg`, `iproute`, `conntrack`, the man pages you link already cover it.

Would you be open to taking it that way? The index is worth having -- I'd just rather it come from the collectors than a doc we keep in sync by hand.

PR #121, opened 2026-04-01
Add expected NVIDIA GPU count validation to DeviceCount check

pjanisze98:
**Issue #, if available:**

  N/A — discovered during investigation of a live production incident.

  **Description of changes:**

  **Problem**

  We discovered a `g6e.12xlarge` EC2 instance running with only 3 of its 4 NVIDIA L40S GPUs visible to the OS. The 4th GPU silently failed to enumerate on the PCIe bus at boot. There were no Xid or dmesg errors and no AWS status check failure. The instance appeared healthy by every existing signal, but was operating at 75% GPU capacity.

  _Why the current check doesn't catch this:_

  The existing DeviceCount check in `dcgm_devices.go` compares:
  - DCGM device count (via NVML GetDeviceCount())
  - /dev/nvidia* file count

  Both of these reflect what the driver sees at the OS level. When a GPU fails to enumerate on the PCIe bus, it never appears to the driver at all, so both DCGM and /dev agree on the (wrong) lower count. The check passed because there's no mismatch between the two.

  **Fix**

  This PR adds a second comparison in DeviceCount() that checks the detected GPU count against the expected count for the EC2 instance type. The expected count is resolved by::
  1. Querying IMDS for the instance type
  2. Looking up the NVIDIA GPU count from an embedded `instance-info.jsonl` lookup table (generated via `hack/generate-instance-info/`)

  If `detected < expected`, a new `NvidiaExpectedDeviceCountMismatch` condition is fired with `Fatal` severity, which sets `AcceleratedHardwareReady=False` on the node.

  The result is cached after the first successful call (since instance type never changes), and failures are retried on the next 5-minute tick rather than permanently cached.
  
**Testing Done:**

  Unit tests

  All 22 tests pass (including 3 new), full build succeeds:

```
$ go vet ./monitors/nvidia/dcgm/... ./pkg/reasons/...
PASS

$ go test -v ./monitors/nvidia/dcgm/...
--- PASS: TestDeviceCount/DeviceCountError
--- PASS: TestDeviceCount/IgnoreNotInitialized
--- PASS: TestDeviceCount/GetDeviceCounts
--- PASS: TestDeviceCountExpectedMismatch/DetectsMissingGPU
--- PASS: TestDeviceCountExpectedMismatch/NoConditionWhenCountMatches
--- PASS: TestDeviceCountExpectedMismatch/HandlesProviderError
PASS
```

  **Live instance verification**

  Verified the detection gap on a production g6e.12xlarge where 1 of 4 GPUs failed to enumerate. The following commands demonstrate that IMDS + EC2 API provide the expected count,
  while the OS only sees 3:
```
$ curl -s -H "X-aws-ec2-metadata-token: $TOKEN" \
    http://169.254.169.254/latest/meta-data/instance-type
g6e.12xlarge

$ aws ec2 describe-instance-types \
    --region us-east-1 \
    --instance-types g6e.12xlarge \
    --query 'InstanceTypes[0].GpuInfo.Gpus[*].{Name:Name,Count:Count}' \
    --output table
-----------------------
|DescribeInstanceTypes|
+----------+----------+
|   Count  |  Name    |
+----------+----------+
|  4       |  L40S    |
+----------+----------+

$ nvidia-smi --query-gpu=index,name,pci.bus_id --format=csv,noheader
0, NVIDIA L40S, 00000000:38:00.0
1, NVIDIA L40S, 00000000:3A:00.0
2, NVIDIA L40S, 00000000:3C:00.0

$ ls /dev/nvidia[0-9]* | wc -l
3

$ sudo dmesg | grep -i -E "xid|nvrm|pci.*error|gpu"
[    5.011741] NVRM: loading NVIDIA UNIX Open Kernel Module for x86_64  570.172.08  Release Build ...
```

Description of the node's capacity
```
$ kubectl describe node {{bad_node}}
Name:               {{bad_node}}
...
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=g6e.12xlarge
                    beta.kubernetes.io/os=linux
...
...
Capacity:
  cpu:                    48
  ephemeral-storage:      3708861428Ki
  hugepages-1Gi:          0
  hugepages-2Mi:          10562Mi
  memory:                 390836892Ki
  nvidia.com/l40s:        3 <---------- THIS SHOULD BE 4
  pods:                   292
  vpc.amazonaws.com/efa:  1
```

  No Xid or PCIe errors, the 4th GPU seems to never appeared on the bus. The current `DeviceCount` check sees DCGM=3 and /dev=3 and reports no issue. With this change, the agent would detect expected=4, actual=3 and mark the node unhealthy.


mselim00 (commented):
thanks for putting this together!

monitors/nvidia/dcgm/instance_type.go
```diff
@@ -0,0 +1,91 @@
...
+	return &ec2ExpectedGPUCountProvider{}
+}
+
+type ec2ExpectedGPUCountProvider struct {
+	mu       sync.Mutex
+	count    uint
+	resolved bool
+}
+
+func (p *ec2ExpectedGPUCountProvider) GetExpectedGPUCount(ctx context.Context) (uint, error) {
```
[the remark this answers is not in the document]
pjanisze98: Done - added an embedded `instance-info.jsonl` with all known GPU instance types, this gets checked first before falling back to the EC2 API

docs/node-health-issues.adoc
mselim00: I think you intended to undo this?

mselim00 (commented):
thanks for the revision!

pkg/instanceinfo/instance-info.jsonl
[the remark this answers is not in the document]
prasad0896: We already have a similar script here: https://github.com/aws/eks-node-monitoring-agent/blob/main/hack/update-nvidia-gpu-list.sh

pkg/reasons/reasons.yaml
mselim00, line 1: can you also clean up this and reasons.go?

monitors/nvidia/dcgm/dcgm.go
```diff
@@ -6,20 +6,35 @@ import (
...
+// InstanceTypeInfoProvider, primarily for testing.
+func NewDCGMSystemWithInstanceTypeInfoProvider(dcgmClient DCGM, diagType dcgmapi.DiagType, provider instanceinfo.InstanceTypeInfoProvider) *DCGMSystem {
+	return &DCGMSystem{
+		dcgm:                    dcgmClient,
+		diagType:                diagType,
+		instanceTypeInfoProvider: provider,
+		fieldValueWindow:         5 * time.Minute,
+	}
+}
+
```
mselim00, line 27: nit: `NewDCGMSystem` can call to `NewDCGMSystemWithInstanceTypeInfoProvider` with `instanceinfo.NewInstanceTypeInfoProvider()`

internal/pkg/instanceinfo/provider.go
```diff
@@ -0,0 +1,140 @@
...
+
+	info := &InstanceInfo{InstanceType: instanceType}
+
+	if gpuInfo := resp.InstanceTypes[0].GpuInfo; gpuInfo != nil {
+		for _, gpu := range gpuInfo.Gpus {
+			if gpu.Count != nil {
+				info.GPUCount += uint(*gpu.Count)
+			}
+		}
+	}
```
mselim00: we should also retain the GPU manufacturer - in our current use, we only want Nvidia GPUs.

take `g4ad.2xlarge` for example - it has two AMD Radeon Pro GPUs. our Nvidia GPU checks will always show 0 for it, which would diverge from the 2 "expected"

mselim00 (commented):
just some more minor changes/cleanup and then lgtm

pjanisze98:
Bump on this! I got busy so wasn't able to get the last set of commits out for a few weeks, but I hope I got everything resolved @mselim00

pjanisze98:
friendly bump on this another time @mselim00 @prasad0896

prasad0896:
@pjanisze98 sorry about the delay. I am taking a look at this PR now.

prasad0896 (commented):
Thank you for the patient work on this PR. We really appreciate it. Some high level feedback before we approve the code change:

* Drop the runtime `ec2:DescribeInstanceTypes` fallback. We've intentionally avoided taking a runtime dependency on regional AWS APIs from the agent - every existing AWS interaction in the agent is either link-local IMDS (which doesn't require IAM) or goes through the Kubernetes API. The fallback would be the agent's first regional AWS API call at runtime, and it also won't work on the EKS Auto Mode. So on Auto Mode the fallback would silently fail every time on any instance type is not present in the JSONL. The embedded JSONL is the right primary path; please make it the only runtime path.

* Update the PR title and description to match the code. The title says `NvidiaExpectedDeviceCountMismatch` and the description says "a new condition is fired," but the diff reuses the existing NvidiaDeviceCountMismatch reason and only differentiates by Message text. Keeping one reason is fine - both failure modes share the same repair action - but the title and description need to match. Additionally, reason codes are customer contract via nodeRepairConfigOverrides, so the same sentence belongs in node-health-issues.adoc

* Race condition on the cache, error-swallowing in JSONL parse, and a few smaller things — see inline comments.

Out of scope for this PR - I'll handle separately:

* The new Go scanner overlaps with the existing update-nvidia-gpu-list.sh(both walk ec2:DescribeInstanceTypes for NVIDIA GPU instance types). Going forward we'll consolidate by replacing the shell script with this Go scanner. No action needed from you on consolidation in this PR.
* I'll also add the GitHub Actions workflow to refresh instance-info.jsonl in a separate follow-up PR. You don't need to add it here.

internal/pkg/instanceinfo/provider.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Try embedded lookup first — no API call needed.
+	if info, ok := p.embeddedLookup[instanceType]; ok {
+		p.info = &info
+		return p.info, nil
+	}
+
+	// Fall back to EC2 API for unknown instance types.
+	info, err := getInstanceInfoFromEC2API(ctx, instanceType)
```
prasad0896: As discussed above, please remove this fallback path entirely. Replace the fallback with a sentinel error so the caller can distinguish "unknown instance type" from "IMDS failed":

internal/pkg/instanceinfo/provider.go
```diff
@@ -0,0 +1,135 @@
...
+// DescribeInstanceTypes API. The result is cached after the first successful
+// resolution.
+func NewInstanceTypeInfoProvider() *ec2InstanceTypeInfoProvider {
+	return &ec2InstanceTypeInfoProvider{
+		embeddedLookup: loadEmbeddedInstanceInfo(),
+	}
+}
+
+type ec2InstanceTypeInfoProvider struct {
+	info           *InstanceInfo
```
prasad0896, line 50: The cache field info is read and written without synchronization. Today DeviceCount is the only caller from a single goroutine so the race is theoretical, but the InstanceTypeInfoProvider interface is exported and a future second caller (e.g. a neuron-equivalent check) would race. Please add a sync.RWMutex and protect the cache read/write.

monitors/nvidia/dcgm/dcgm.go
prasad0896: +1 to above

pjanisze98:
Updated PR summary and title to reflect current design.

Added a sync.RWMutex to protect the cache

removed the `getInstanceInfoFromEC2API()` fallback. This only uses the jsonl now.

prasad0896:
thanks for the changes and the PR lgtm.

prasad0896 (approved):

PR #153, opened 2026-05-01
feat: add InstanceStoreIOError detection to storage monitor

rcipdev:
**Issue #, if available**:

**Description of changes**:
adds detection of block device I/O errors from dmesg to the storage monitor. The agent now matches three kernel log patterns (end_request, Buffer I/O error, blk_update_request) and reports a Warning BlockDeviceIOError condition on StorageReady. Warning severity is used for the initial release so the issue surfaces as a node Event without triggering Auto Repair.
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/TroubleshootingInstances.html#DeviceBlock

**Testing Done**:
built locally and tested with fatal condition on 1.33 cluster
`{"level":"info","ts":"2026-05-01T22:43:17Z","msg":"sending condition to exporter","hostname":"...","condition":{"Reason":"InstanceStoreIOError","Message":"I/O error detected on block device nvme1n1 at sector 12345678, indicating a failed physical drive or EBS volume","Severity":"Fatal","MinOccurrences":0},"conditionType":"StorageReady"}`


docs/node-health-issues.adoc
```diff
@@ -349,6 +349,10 @@ The monitoring condition is `StorageReady` for issues in the following table tha
 |Event
 |Input or output delay detected in a process, potentially indicating insufficient input-output provisioning if excessive.
 
+|InstanceStoreIOError
```
prasad0896: Should this be `BlockDeviceIOError`?

pkg/reasons/reasons.go
```diff
@@ -288,6 +288,10 @@ var (
         template:        "IODelays",
         defaultSeverity: "Warning",
     }
+    InstanceStoreIOError = ReasonMeta{
+        template:        "InstanceStoreIOError",
+        defaultSeverity: "Fatal",
```
prasad0896: A single kernel I/O error log triggers StorageReady=False, which Auto Repair acts on. But block device I/O errors in dmesg are emitted in many situations that are not permanent hardware failures: brief NVMe controller hiccups, transient EBS unavailability or throttling boundaries, userspace bugs touching a stale device, even some live-migration paths. Compare with sibling reasons in this same file:

    - ExcessiveZombieProcesses uses MinOccurrences(5) → 25 minutes of exposure before flipping.
    - IODelays uses a per-PID delay cache with a >10s threshold.
    - RapidCron requires repeated observation across a window.

Recommendation: Either drop severity to Warning for the initial release, or set a meaningful MinOccurrences (≥3)

prasad0896:
@rcipdev thanks for changes. 

`MinOccurrences` counts occurrences, not elapsed time, and the manager only resets the counter after the condition fires. That works for the poll-based reasons (they're re-evaluated on a timer, so the count reflects sustained duration), but this handler emits once per matching dmesg line. A real block-device fault prints many error lines back-to-back, so MinOccurrences(3) is crossed almost instantly - and because there's no time-based reset, occasional unrelated I/O errors also accumulate and eventually trip it. Net effect: it reads like a debounce but doesn't actually prevent a transient burst from flipping StorageReady=False and triggering Auto Repair. For a Fatal signal we'd want either a true rate threshold (N errors within T minutes) or Warning severity for the initial release.

prasad0896 (changes requested):

monitors/storage/monitor_test.go
```diff
@@ -153,3 +153,72 @@ func TestPeriodic(t *testing.T) {
...
+			shouldDetect: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var matched bool
+			var device, location string
+
+			if matches := ioErrorEndRequest.FindStringSubmatch(tt.logLine); len(matches) > 2 {
```
prasad0896: This test should test this function: `handleBlockDeviceIOErrors` instead of re-implementing the same logic in test.

monitors/storage/monitor.go
```diff
@@ -143,6 +149,42 @@ func (m *StorageMonitor) handleKubeletLogs(line string) error {
 	return nil
 }
 
+// Regex patterns for detecting I/O errors on block devices
+// Matches patterns like:
+// - "end_request: I/O error, dev sde, sector 52428288"
+// - "Buffer I/O error on device md0, logical block 209713024"
+// - "blk_update_request: I/O error, dev nvme1n1, sector 12345"
+var (
+	ioErrorEndRequest = regexp.MustCompile(`end_request: I/O error, dev ([a-z0-9]+), sector ([0-9]+)`)
```
prasad0896: We should update the device regex to `[a-z0-9-]+` so as to catch device names like `dm-0`

prasad0896 (changes requested):

prasad0896 (approved):

Issue #158, opened 2026-05-08
Startup CrashLoopBackOff on freshly-provisioned nodes — probe not tunable, default cpu limit not omittable


ikibko3c:
## Summary

`eks-node-monitoring-agent` pods (addon `v1.6.4-eksbuild.1`, chart 1.6.4) restart twice on every freshly-provisioned, contention-heavy node before stabilizing. This trips `KubePodCrashLooping` alerts in any monitoring stack that uses the standard kube-prometheus rules. Two compounding chart limitations make it impossible to fix the root cause through chart values or addon `configurationValues`.

## Reproduction

Cluster with Karpenter provisioning new nodes for bursty workloads (CI runners, etc.). When a node comes up with several pods scheduling at once, the agent's liveness probe fails:

```
Warning  Unhealthy  Liveness probe failed:
  Get "http://<podIP>:8002/healthz": context deadline exceeded
Normal   Killing    Container failed liveness probe, will be restarted
```

The pod is killed twice (Exit 143 / SIGTERM) at exactly 30s each (3 probe failures × 10s period), then stabilizes. Confirmed on multiple Karpenter spot pools (m6g.xlarge / m6g.2xlarge ARM nodes), reproducible on every fresh node, never on long-running nodes.

## Root cause

Two separate issues, both in the chart:

### 1. Liveness probe timing is not templatable

`charts/eks-node-monitoring-agent/templates/daemonset.yaml` renders only the port:

```yaml
livenessProbe:
  httpGet:
    path: /healthz
    port: {{ .Values.nodeAgent.probePort | default 8002 }}
```

Kubernetes therefore applies defaults: `timeoutSeconds: 1, periodSeconds: 10, failureThreshold: 3, initialDelaySeconds: 0`. The agent takes ~15s to initialize all four monitors (kernel/storage/networking/runtime), and on busy nodes the `/healthz` handler can't respond within the 1s timeout under CPU contention. There is no `startupProbe`, no `initialDelaySeconds`, and no exposed timing knob.

The agent's CLI (`cmd/eks-node-monitoring-agent/main.go`) has no probe-related flags either, so `additionalArgs` cannot mitigate this.

### 2. CPU limit cannot be omitted via configurationValues

The natural workaround — remove the default `250m` CPU limit so CFS doesn't throttle the healthz handler — is impossible:

| Attempt | Result |
|---|---|
[...]

PR #159, opened 2026-05-08
feat(chart): make liveness/startup probes configurable, raise default cpu limit

ikibko3c:
## Summary

Fixes the startup race documented in #158: `eks-node-monitoring-agent` pods restart twice on every freshly-provisioned, contention-heavy node before stabilizing, producing `KubePodCrashLooping` alerts.

Three coordinated changes, all backwards-compatible:

1. **Expose `nodeAgent.livenessProbe` timing in chart values** (`templates/daemonset.yaml`, `values.yaml`). Empty default `{}` preserves current k8s probe defaults — no behavior change for existing users. Operators can now set `timeoutSeconds`, `failureThreshold`, etc. without forking the chart.
2. **Add optional `nodeAgent.startupProbe`**. Empty default `{}` means no startup probe is rendered; opting in lets operators absorb the monitor-init burst without loosening the steady-state liveness probe. Path and port are derived from `probePort` for consistency.
3. **Raise default `nodeAgent.resources.limits.cpu` from `250m` to `500m`**. Agent steady-state usage is ~5–10m; the prior limit caused CFS throttling during parallel monitor initialization, which is the actual root cause of the `/healthz` timeouts. Memory limit is unchanged. This is the only change that affects existing users — and it makes things strictly better for the most common failure mode.

`charts/configuration.schema.json` is updated with a `Probe` definition referenced from both new fields, so EKS addon `configurationValues` can pass them through validation.

## Render verification

Default values (no override) — startupProbe correctly absent, liveness gets k8s defaults:

```yaml
livenessProbe:
  httpGet:
    path: /healthz
    port: 8002
resources:
  limits:
    cpu: 500m
    memory: 200Mi
  requests:
    cpu: 10m
    memory: 30Mi
```

With `nodeAgent.livenessProbe.timeoutSeconds=5`, `nodeAgent.startupProbe.failureThreshold=12`:

```yaml
livenessProbe:
  httpGet:
    path: /healthz
    port: 8002
  timeoutSeconds: 5
startupProbe:
  httpGet:
    path: /healthz
    port: 8002
  failureThreshold: 12
```

## Test plan

- [x] `helm lint ./charts/eks-node-monitoring-agent` passes
- [x] `helm template` renders correct output for default and override cases (above)
- [x] `make helm-docs` regenerates README cleanly
- [x] `configuration.schema.json` is valid JSON and `Probe` is reachable from `NodeAgent.{livenessProbe,startupProbe}`
- [ ] CI checks on this PR

## Backwards compatibility

- Default render of `livenessProbe` is byte-identical to current main (empty `{}` produces no merged fields).
- `startupProbe` is not rendered when empty, so no new probe appears for users who don't opt in.
- `resources.limits.cpu` default change is the one observable diff. Operators who care can pin via values or `configurationValues`.

Refs #158

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/charts/configuration.schema.json
+++ b/charts/configuration.schema.json
@@ -51,6 +51,14 @@
                     "description": "Health probe port for the eks node monitoring agent. Used for both the --probe-address arg and the liveness probe.",
                     "default": 8002
                 },
+                "livenessProbe": {
+                    "$ref": "#/definitions/Probe",
+                    "description": "Liveness probe timing for the eks node monitoring agent. Path and port are derived from probePort; supply timing fields to override Kubernetes defaults."
+                },
+                "startupProbe": {
+                    "$ref": "#/definitions/Probe",
+                    "description": "Optional startup probe for the eks node monitoring agent. When unset, no startup probe is rendered. Path and port are derived from probePort."
+                },
                 "resources": {
                     "$ref": "#/definitions/Resources"
                 },
@@ -157,6 +165,34 @@
                 },
                 "cpu": {
                     "type": "string"
+                }
+            }
+        },
+        "Probe": {
+            "title": "Probe",
+            "type": "object",
+            "description": "Probe timing fields. Path and port are derived from probePort.",
+            "additionalProperties": false,
+            "properties": {
+                "initialDelaySeconds": {
+                    "type": "integer",
+                    "minimum": 0
+                },
+                "timeoutSeconds": {
+                    "type": "integer",
+                    "minimum": 1
+                },
+                "periodSeconds": {
+                    "type": "integer",
+                    "minimum": 1
+                },
+                "successThreshold": {
+                    "type": "integer",
+                    "minimum": 1
+                },
+                "failureThreshold": {
+                    "type": "integer",
+                    "minimum": 1
                 }
             }
         },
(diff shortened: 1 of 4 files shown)


prasad0896 (changes requested):
@ikibko3c thanks for digging into this — exposing probe timing and bumping the CPU limit are reasonable mitigations, and the chart work itself is clean.

Stepping back on #158, though: probe timing treats the symptom rather than the cause. The root cause is that the agent's `/healthz` server doesn't start listening until `mgr.Start()`, which is the **last** call in `run()` (`cmd/eks-node-monitoring-agent/main.go`) — after the synchronous startup path completes (Kubernetes client setup, a blocking node-bootstrap poll, and all monitor initialization). So for the first ~15s nothing answers on `:8002`; a startupProbe just widens how long we wait, but the endpoint is still dark during init, and any node slow enough still trips it.

The fix at the source is to get the health server listening from ~t0: reach `mgr.Start()` early and move the blocking startup work into manager `Runnable`s (`mgr.Add(...)`) that the manager starts *after* the probe server is up. controller-runtime starts the health probe server early in `Start()` (before cache sync), and `healthz.Ping` returns 200 immediately — so the liveness probe passes from the start and the crashloop can't occur regardless of how long initialization takes. There's precedent for exactly this in the repo: #152 moved the NodeDiagnostic informer registration into a `manager.RunnableFunc` to avoid a blocking call before `Start()`.

Would you be open to taking the PR in that direction? A default startupProbe is fine to keep as defense-in-depth, but the reordering is what actually resolves #158 at the source. Happy to review along the way.

prasad0896:
We also need to fix the auto-generation and conflicts in this PR.

prasad0896:
We have merged #212 which should fix all the issues with startup. We can add the ability to configure startup and liveness if we continue to see issues even after #212 is released. Feel free to reopen if it does not address your problem.

PR #163, opened 2026-05-19
fix(networking): migrate CRI client to k8s.io upstream

prasad0896:
**Issue #, if available**:

**Description of changes**:

The networking monitor was importing github.com/containerd/containerd/v2/integration/{cri-api/pkg/apis,remote}. Those paths live inside containerd's integration test tree and carry no API stability commitment. containerd v2.3.0 deleted them as part of adopting k8s.io/cri-client upstream, which broke the weekly update-go-deps workflow with:
```
    module github.com/containerd/containerd/v2@latest found (v2.3.0),
    but does not contain package
    github.com/containerd/containerd/v2/integration/cri-api/pkg/apis
```
Switch to the canonical SIG Node packages that containerd itself now depends on:
```
    - cri "k8s.io/cri-api/pkg/apis"          (interface types)
    - criclient "k8s.io/cri-client/pkg"      (gRPC client constructor)
```
These are versioned with Kubernetes, follow the standard deprecation policy, and are independent of the container runtime's release cadence. Future containerd bumps no longer churn this code.

The new client's PodSandboxStatus signature gained a context.Context and a verbose bool, and returns *PodSandboxStatusResponse rather than *PodSandboxStatus. Field reads are updated accordingly. The new NewRemoteRuntimeService takes optional tracer and logger arguments (passed as nil here) and eagerly validates the CRI connection at construction time rather than deferring to first call.

Add a WithRuntimeService Option so checkIPAMD can be exercised under test without dialing a real CRI socket, and add a unit test that verifies the IPAMDInconsistentState path end-to-end against a fake runtime service. The pre-existing TestCheckIPAMD_CacheExpiry now injects the fake as well, since the new client's eager validation would otherwise fail in environments without /run/containerd/containerd.sock.
```
  go.mod: drop github.com/containerd/containerd/v2 (no longer imported),
  promote k8s.io/cri-api from indirect to direct, add k8s.io/cri-client.
```

**Testing Done**:
Added unit tests for this change


junpengdev:
Looks good to me but the dependency review failed.

junpengdev (approved):

PR #164, opened 2026-05-20
fix(nvidia): emit FabricManagerNotRunning Fatal when FM has failed

prasad0896:
**Issue #, if available**:

**Description of changes**:
The fabric-field handler in WatchFields was sitting behind the DCGM_ST_OK/DCGM_ST_NO_DATA early-continue gate. DCGM returns DCGM_ST_OK for a successfully-read field even when the field's value indicates failure (FM status = 4 means the fabricmanager service is in Failed state), so the handler was never invoked in production and FabricManagerNotRunning was never emitted.

Move handleFabricField above the OK/NO_DATA gate. Update the fabric subtests to use DCGM_ST_OK so they exercise the production-realistic path; the previous DCGM_ST_BADPARAM setup masked the bug.

**Testing Done**:


junpengdev (approved):

prasad0896:
CI tests passed but the cluster deletion post tests failed. So this is good to be merged.

Issue #99, opened 2026-03-18
feat: add agent version info to logs


youwalther65:
**What would you like to be added**:
In Auto Mode it is not possible to determine the underlying NMA version.
When I ran a debug container to attach to the node I just got:
```bash
$ kubectl debug node/i-0010eb088fa55e59d -it --profile=sysadmin --image=public.ecr.aws/amazonlinux/amazonlinux:2023
...
No version information:
bash-5.2# nsenter -t 1 -m journalctl -u eks-node-monitoring-agent | more
Mar 18 06:07:49 ip-192-168-113-254.eu-west-1.compute.internal systemd[1]: Starting eks node monitoring agent...

# No output
bash-5.2# nsenter -t 1 -m journalctl -u eks-node-monitoring-agent | grep -i version
```

Providing this info in logs means, the version is available in Auto Mode and in OSS version in container logs as well.
**Why is this needed**:
In Auto Mode NMA is running as a systemd unit and it is not possible to determine the version.
Check what features are available by comparing version with OSS GitHub repo releasees.

**Are you currently working around this issue?**:
N/A

How are you currently solving this problem?
N/A

prasad0896:
@youwalther65 we have merged #173 that should allow publishing the versions in the Auto AMIs starting next week. Let me know if there is anything else you need before we close out this issue.

PR #173, opened 2026-06-09
feat: add agent version info during build time

prasad0896:
**Issue #, if available**:
#99 

**Description of changes**:

The agent doesn't log its own version anywhere, so in Auto Mode there's no way to tell which build is running on a node when you're triaging. This wires that in.

Version and commit now live in a small internal/version package, set at link time with -ldflags -X. The Makefile pulls the version from the chart's nodeAgent.image.tag and falls back to git describe, and the commit comes from git rev-parse; both flow through the Dockerfile build args. main.go logs them on the first line at startup. Without the ldflags an ordinary go build still works, it just reports build (unknown).

**Testing Done**:
* Added unit tests for the change
* Built the code changes binary and validated against the latest shipped code:
```
strings bin/eks-node-monitoring-agent → v1.6.5-eksbuild.1
                                       → 6a989b4
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,18 @@ IMAGE_TAG ?= latest
 # Docker build arguments
 GOBUILDARGS ?=
 
+# Build identity (stamped into the binary via -ldflags -X).
+# VERSION resolves from the first available source:
+#   1. the chart's nodeAgent.image.tag (the canonical released version, and the
+#      same value the release pipeline uses to tag the image) — works in build
+#      environments without git metadata (e.g. Brazil),
+#   2. git describe for local/OSS checkouts,
+#   3. "build" as a final fallback.
+CHART_VALUES    := charts/eks-node-monitoring-agent/values.yaml
+CHART_IMAGE_TAG := $(shell yq '.nodeAgent.image.tag' $(CHART_VALUES) 2>/dev/null)
+VERSION    ?= $(if $(CHART_IMAGE_TAG),$(CHART_IMAGE_TAG),$(shell git describe --tags --always --dirty 2>/dev/null || echo build))
+GIT_COMMIT ?= $(shell git rev-parse --short HEAD 2>/dev/null || echo unknown)
+
 # Multi-arch build platforms
 DOCKER_PLATFORMS ?= linux/amd64,linux/arm64
 
@@ -107,9 +119,12 @@ help: ## Show this help message
 # Development Targets
 # =============================================================================
 
+_version_pkg := github.com/aws/eks-node-monitoring-agent/internal/version
+LDFLAGS := -X $(_version_pkg).Version=$(VERSION) -X $(_version_pkg).GitCommit=$(GIT_COMMIT)
+
 .PHONY: build
 build: generate fmt vet ## Build Go code
-	go build -o $(OUTPUT_BIN)/eks-node-monitoring-agent ./cmd/eks-node-monitoring-agent
+	go build -ldflags "$(LDFLAGS)" -o $(OUTPUT_BIN)/eks-node-monitoring-agent ./cmd/eks-node-monitoring-agent
 	go build -o $(OUTPUT_BIN)/chroot ./cmd/chroot
 
 .PHONY: test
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.26.2
 
 require (
 	github.com/NVIDIA/go-dcgm v0.0.0-20260603204728-453d82102783
-	github.com/aws/amazon-vpc-cni-k8s v1.22.1
```
prasad0896, line 7: Intentional downgrade because of this: https://github.com/aws/amazon-vpc-cni-k8s/issues/3721

internal/version/version.go
```diff
@@ -0,0 +1,17 @@
...
+
+// GitCommit is the short git SHA the binary was built from.
+// Default "unknown" indicates an unstamped local build.
+var GitCommit = "unknown"
+
+// String returns the human-readable identity string in the form
+// "<Version> (<GitCommit>)".
+func String() string {
+	return Version + " (" + GitCommit + ")"
+}
```
shvbsle, line 19: Looks like dead code?

Also a suggestion:
```suggestion
func String() string {
    return fmt.Sprintf("%s (%s)", Version, GitCommit)
}
```

cmd/eks-node-monitoring-agent/main.go
```diff
@@ -104,6 +105,11 @@ func run() error {
 	logger := zap.New(zap.Level(zapraw.NewAtomicLevelAt(zapcore.Level(-verbosity)))).WithValues("hostname", hostname)
 	log.SetLogger(logger)
 
+	logger.Info("starting eks-node-monitoring-agent",
+		"version", version.Version,
```
shvbsle: Did you intend to use the `String()` method here? otherwise String() seems like dead-code

Makefile
```diff
@@ -107,9 +119,11 @@ help: ## Show this help message
 # Development Targets
 # =============================================================================
 
+LDFLAGS := -X github.com/aws/eks-node-monitoring-agent/internal/version.Version=$(VERSION) -X github.com/aws/eks-node-monitoring-agent/internal/version.GitCommit=$(GIT_COMMIT)
```
shvbsle:
```suggestion
_version_pkg := github.com/aws/eks-node-monitoring-agent/internal/version
LDFLAGS := -X $(_version_pkg).Version=$(VERSION) -X $(_version_pkg).GitCommit=$(GIT_COMMIT)
```

Makefile
```diff
@@ -22,6 +22,18 @@ IMAGE_TAG ?= latest
...
 
+# Build identity (stamped into the binary via -ldflags -X).
+# VERSION resolves from the first available source:
+#   1. the chart's nodeAgent.image.tag (the canonical released version, and the
+#      same value the release pipeline uses to tag the image) — works in build
+#      environments without git metadata (e.g. Brazil),
+#   2. git describe for local/OSS checkouts,
+#   3. "build" as a final fallback.
+CHART_VALUES    := charts/eks-node-monitoring-agent/values.yaml
+CHART_IMAGE_TAG := $(shell awk '/^[[:space:]]*tag:/ {print $$2; exit}' $(CHART_VALUES) 2>/dev/null)
```
shvbsle: Wondering if we can just use `yq` instead to parse through the yaml files.
```bash
CHART_IMAGE_TAG := $(shell yq '.nodeAgent.image.tag' \
    charts/eks-node-monitoring-agent/values.yaml 2>/dev/null)
```

We already use it in this repo in couple of places [example](https://github.com/aws/eks-node-monitoring-agent/blob/6a989b4030c62da63eb457b5946b24d00c237b3e/hack/update-nvidia-gpu-list.sh#L40)

Makefile
prasad0896: Thanks for the suggestion, will update

Makefile
prasad0896: yup, can swap this for yq

shvbsle (approved):

youwalther65:
Sounds great, thank you @prasad0896 !

PR #177, opened 2026-06-10
Collect NPA metric and report as Node condition

Issacwww:
## Problem

NPA (Network Policy Agent) crash-looped for ~18h undetected because NMA had no NPA-specific health monitoring. This adds NPA health detection on EKS
Auto Mode nodes, surfaced through NMA's existing pipeline under `NetworkingReady`. The checks live in the networking monitor (peer to the IPAMD checks), so they inherit the `NetworkingReady` condition type.

## What this adds

| Signal | Source | Severity | Surfaces as | Fires when | Why this severity |
|--------|--------|----------|-------------|-----------|-------------------|
| `NPANotRunning` | D-Bus `ActiveState` | **Fatal** | `NetworkingReady` condition (→ auto-repair) | down ≥ 15 min | Persistently down → terminal; node replacement is the fix. |
| `NPARepeatedlyRestart` | D-Bus `NRestarts` | **Warning** | node Event | ≥ 5 restarts in one poll | Crash loop; root cause often not node-local, so replacing the node may not help → alarm, don't auto-repair. |
| `NPABPFRecoveryError` | NPA log | **Warning** | node Event | ≥ 2 matching error lines | Best-effort startup recovery that self-heals on the next reconcile → not terminal. |

Scope: **Auto Mode only**. Severity mirrors IPAMD's split (persistent-down = Fatal, restart/recovery = Warning).

### Deferred to a follow-up PR

**Policy-sync drift** — the authoritative "policy actually not enforced" signal. It needs NPA to expose desired-vs-loaded policy counts as Prometheus gauges on its existing metrics endpoint; NMA would then scrape them and emit `NPAPolicySyncFailure` (Fatal) on sustained drift. Also deferred: NPA health-endpoint liveness, and moving these signals onto the shared metrics transport (separate effort).

## Testing

**Unit**: detection logic (`checkRepeatedlyRestart`, `checkNotRunning`,
`HandleLogs`) incl. threshold/window edges, and the Auto-only gate.

**On-node** (custom AMI pinned to a beta Auto Mode cluster):
- `NPABPFRecoveryError`: appended matching error lines to the NPA log to mock the
  failure — NMA tailed the log and emitted the Warning. ✅
- `NPARepeatedlyRestart`: SIGKILL'd NPA ≥5× within one poll to force systemd auto-restarts — NMA detected the loop via D-Bus and emitted the Warning. ✅
- `NPANotRunning`: masked NPA and held > 15 min — `NetworkingReady` flipped to  `False` (Fatal). ✅ 
- Negatives: sub-threshold restarts and a short (<15 min) outage produced no
  condition.

**Performance**: not covered. Footprint is small — one D-Bus poll (2 property
reads) every 5 min plus an append-driven log tail — but there is no formal
perf/load test. Calling this out as a known gap.

## Recovery model

- **Warning** (`NPARepeatedlyRestart`, `NPABPFRecoveryError`) → node Event. No state to clear; it just stops firing once the issue goes away. 
- **Fatal** (`NPANotRunning`) → `NetworkingReady=False`. It does not self-clear; recovery is node auto-repair replacing the node (same as IPAMD's Fatal conditions today).

### DIFF
--- a/pkg/reasons/reasons.go
+++ b/pkg/reasons/reasons.go
@@ -249,6 +249,18 @@ var (
         template:        "MissingLoopbackInterface",
         defaultSeverity: "Fatal",
     }
+    NPABPFRecoveryError = ReasonMeta{
+        template:        "NPABPFRecoveryError",
+        defaultSeverity: "Warning",
+    }
+    NPANotRunning = ReasonMeta{
+        template:        "NPANotRunning",
+        defaultSeverity: "Fatal",
+    }
+    NPARepeatedlyRestart = ReasonMeta{
+        template:        "NPARepeatedlyRestart",
+        defaultSeverity: "Warning",
+    }
     NetworkSysctl = ReasonMeta{
         template:        "NetworkSysctl",
         defaultSeverity: "Warning",
--- a/docs/node-health-issues.adoc
+++ b/docs/node-health-issues.adoc
@@ -299,6 +299,18 @@ The monitoring condition is `NetworkingReady` for issues in the following table 
 |Condition
 |The loopback interface is missing from this instance, causing failure of services depending on local connectivity.
 
+|NPABPFRecoveryError
+|Event
+|The Network Policy Agent encountered errors recovering or loading its BPF programs/maps after a restart. This is typically self-healing as the agent re-creates state on the next policy reconcile, so it is surfaced as a Warning.
+
+|NPANotRunning
+|Condition
+|The Network Policy Agent process was not found running on this Auto Mode node for more than 15 minutes.
+
+|NPARepeatedlyRestart
+|Event
+|The Network Policy Agent has restarted 5 or more times within a single poll interval, indicating a crash loop that may disrupt network policy enforcement. Often caused by a non-node-local issue (e.g. bad config or control-plane input), so it is surfaced as a Warning rather than triggering node replacement.
+
 |NetworkSysctl
 |Event
 |This node's network `sysctl` settings are potentially incorrect.
--- a/monitors/networking/monitor.go
+++ b/monitors/networking/monitor.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/eks-node-monitoring-agent/monitors/networking/ipamd"
 	"github.com/aws/eks-node-monitoring-agent/monitors/networking/iptables"
 	"github.com/aws/eks-node-monitoring-agent/monitors/networking/networkutils"
+	"github.com/aws/eks-node-monitoring-agent/monitors/networking/npa"
 	"github.com/aws/eks-node-monitoring-agent/pkg/config"
 	"github.com/aws/eks-node-monitoring-agent/pkg/osext"
 	"github.com/aws/eks-node-monitoring-agent/pkg/reasons"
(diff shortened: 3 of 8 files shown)


monitors/networking/plugin.go
```diff
@@ -2,13 +2,16 @@ package networking
...
+	"github.com/aws/eks-node-monitoring-agent/monitors/networking/npa"
+	"github.com/aws/eks-node-monitoring-agent/pkg/config"
 	"github.com/aws/eks-node-monitoring-agent/pkg/monitor/framework"
 	"github.com/aws/eks-node-monitoring-agent/pkg/monitor/registry"
 )
 
 func init() {
 	plugin := framework.NewPlugin("networking", []monitor.Monitor{
 		NewNetworkingMonitor(),
+		npa.NewNPAMonitor(config.GetRuntimeContext()),
```
prasad0896: Can we move this out of plugin and into the monitor. See how we do for ipamd: https://github.com/aws/eks-node-monitoring-agent/blob/main/monitors/networking/monitor.go#L324

pkg/reasons/reasons.go
```diff
@@ -249,6 +249,22 @@ var (
         template:        "MissingLoopbackInterface",
         defaultSeverity: "Fatal",
     }
+    NPACrashLoop = ReasonMeta{
+        template:        "NPACrashLoop",
+        defaultSeverity: "Warning",
```
prasad0896, line 254: All NPA reasons are `warning` . Are we certain that is expected?

monitors/networking/npa/monitor_test.go
```diff
@@ -0,0 +1,203 @@
...
+				assertNoCondition(t, mgr)
+			}
+		})
+	}
+}
+
+// --- EKSAuto gate ---
+
+func TestRegister_SkipsWithoutEKSAutoTag(t *testing.T) {
+	// The default runtime context in the test environment does not carry the
```
prasad0896: Preferably we do gating similar to this: https://github.com/aws/eks-node-monitoring-agent/blob/main/e2e/suites/monitors/nvidia_monitor.go#L88 . We have other examples in the repository for the similar use cases.

prasad0896 (commented):
Skimmed through the code and left some feedback.

monitors/networking/plugin.go
Issacwww: yes, moved to monitor in rev2

pkg/reasons/reasons.go
Issacwww: updated, move some to Fatal, with reasoning in the description

monitors/networking/npa/monitor_test.go
Issacwww: Added the gating for NPA in the monitor

prasad0896 (approved):
Overall looks good.

One nit: `checkRepeatedlyRestart` takes an unsigned delta off a baseline seeded at 0. Two false positives follow as a result:
* When the agent restarts, the baseline resets to 0 and the next poll reads the unit's lifetime NRestarts — already ≥ threshold on a long-lived node, so it fires for a healthy agent.
*  A systemd counter reset (daemon-reexec/reset-failed) underflows `current - previous` to ~4.2B.

Fix is the seed-first pattern `checkNotRunning` already uses here, plus a reset guard:
```
if previousRestarts == 0 {
    previousRestarts = currentRestarts
    return // seed baseline, skip first evaluation
}
if currentRestarts < previousRestarts {
    previousRestarts = currentRestarts // counter reset; re-seed
    return
}
```

PR #181, opened 2026-06-17
fix(nvidia/dcgm): decode fabric health mask sub-fields to avoid false positives

junpengdev:
The fabric health handler flagged DCGM_FI_DEV_FABRIC_HEALTH_MASK (field 174) as a fault whenever the mask was non-zero. The mask is a packed bitfield of 2-bit sub-fields decoded as `(mask >> shift) & width` (mirroring NVIDIA's DCGM_GPU_FABRIC_HEALTH_TEST macro), where the value encoding is `0=NotSupported, 1=True (fault), 2=False (healthy)`. Non-zero therefore does NOT imply a fault: e.g. mask `0x80` decodes to `access_timeout_recovery=False` (value 2 at shift 6), a fully healthy GPU. Treating any non-zero mask as an error
produced false-positive node disruptions.

Decode each sub-field and flag only the True/fault state, preserving the granular signal the mask provides. The condition message now names the faulting sub-fields (e.g. "GPU fabric health mask 0x10: route_unhealthy=1") to aid triage.

ref: https://docs.nvidia.com/deploy/nvml-api/group__nvmlFabricDefs.html

**Issue #, if available**:

**Description of changes**:

**Testing Done**:

### DIFF
--- a/monitors/nvidia/dcgm/dcgm_watchfield.go
+++ b/monitors/nvidia/dcgm/dcgm_watchfield.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	dcgmapi "github.com/NVIDIA/go-dcgm/pkg/dcgm"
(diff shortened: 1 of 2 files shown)


monitors/nvidia/dcgm/dcgm_watchfield.go
```diff
@@ -143,6 +156,17 @@ const (
 	DcgmFMStatusSuccess      int64 = 3
 )
 
+// Fabric health summary values for DCGM_FI_DEV_FABRIC_HEALTH_SUMMARY (field
+// 175), from NVML_GPU_FABRIC_HEALTH_SUMMARY_*. go-dcgm only carries these as
+// header comments, so they are redeclared here.
+// ref: https://docs.nvidia.com/deploy/nvml-api/group__nvmlFabricDefs.html (NVML_GPU_FABRIC_HEALTH_SUMMARY_*)
+const (
+	DcgmFabricHealthSummaryNotSupported    int64 = 0
```
shvbsle: These constants are declared but not used

monitors/nvidia/dcgm/dcgm_watchfield.go
```diff
@@ -99,13 +99,26 @@ func handleFabricField(fv dcgmapi.FieldValue_v2) (*monitor.Condition, bool) {
...
+		// access_timeout_recovery=False), so treating a non-zero mask as a
+		// fault produced false positives.
+		// ref: https://docs.nvidia.com/deploy/nvml-api/group__nvmlFabricDefs.html (NVML_GPU_FABRIC_HEALTH_SUMMARY_*)
+		if fv.Int64() != DcgmFabricHealthSummaryUnhealthy {
 			return nil, true
 		}
 		c := reasons.NvidiaFabricError.
 			Builder().
-			Message(fmt.Sprintf("GPU fabric health mask: 0x%x", fv.Int64())).
+			Message("GPU fabric health summary: Unhealthy").
```
shvbsle: Can we enrich this message?

Per https://docs.nvidia.com/datacenter/dcgm/latest/dcgm-api/dcgm-api-field-ids.html#c.DCGM_FI_DEV_FABRIC_HEALTH_SUMMARY

We have 4 possible states for this fabric summary:
```
NVML_GPU_FABRIC_HEALTH_SUMMARY_NOT_SUPPORTED (0)

NVML_GPU_FABRIC_HEALTH_SUMMARY_HEALTHY (1)

NVML_GPU_FABRIC_HEALTH_SUMMARY_UNHEALTHY (2)

NVML_GPU_FABRIC_HEALTH_SUMMARY_LIMITED_CAPACITY (3) 
```


A message like this would be useful in triaging:

```
GPU fabric health summary: Unhealthy with code {NVML_GPU_FABRIC_HEALTH_SUMMARY_UNHEALTHY}
```
(there can be better message formats)

monitors/nvidia/dcgm/dcgm_watchfield.go
```diff
@@ -99,13 +99,26 @@ func handleFabricField(fv dcgmapi.FieldValue_v2) (*monitor.Condition, bool) {
 			Message(fmt.Sprintf("Fabric Manager status: %s", name)).
 			Build()
 		return &c, true
-	case dcgmapi.DCGM_FI_DEV_FABRIC_HEALTH_MASK:
```
shvbsle, line 102: The mask is actually very useful signal. I think instead of moving away from the mask, can we not improve or make this logic more robust?

prasad0896: I agree with @shvbsle that we must keep watching the raw mask (`DCGM_FI_DEV_FABRIC_HEALTH_MASK`). The summary field (175) obscures the specific root cause of the fabric failure, such as a training failure or an access timeout. We need those details for triage.

We can fix the evaluation logic instead of switching fields. We need a Go helper function mimicking NVIDIA's `DCGM_GPU_FABRIC_HEALTH_TEST` macro. We iterate over the mask in 2-bit chunks  `((mask>> shift) & 0x3)` and flag an error only if a chunk evaluates to  1  (True/Fault). This eliminates false positives and retains granular error signals. Let's update the logic to decode the mask properly.

prasad0896 (changes requested):

monitors/nvidia/dcgm/dcgm_watchfield.go
junpengdev: Now it decodes 174 field to flag real faults.

prasad0896 (approved):

anupsinghfra:
This fix prevents a nasty class of false positives that would be extremely hard for
  customers to debug. Imagine tracking down why your GPU nodes keep getting replaced
  when nvidia-smi shows everything green - the DCGM mask would look like noise without
  understanding the encoding.

  The architectural lesson here: when we're surfacing hardware telemetry through
  Kubernetes primitives (NodeConditions), we need to match the signal fidelity of the
  source. Great work preserving the diagnostic value while eliminating the false alerts.

PR #186, opened 2026-06-27
feat: allow configuring the name of the cni daemonset

grosser:
**Description of changes**:

got this error but we have cni ... but it's named `aws-cni`
```
{"level":"info","ts":"2026-06-27T17:33:52Z","msg":"Skipping MAC address policy check - VPC CNI is not installed","hostname":"ip-172-16-112-103.us-west-2.compute.internal","monitor":"networking"}
```

renaming it would be a pain (risk of downtime when we delete so need to slowly migrate with nodeselector hackery)
would this env var be acceptable ?


grosser:
still useful but we don't need it anymore ... FYI the trick is to set `aws.cluster.x-k8s.io/prevent-deletion` and rename it

PR #190, opened 2026-06-30
fix: avoid nil panic for kubeconfigs without an exec credential plugin

grosser:
podRestConfigProvider.Provide() unconditionally dereferenced restConfig.ExecProvider to rewrite the exec command/args so the IAM authenticator runs against host binaries. This assumes the host kubeconfig always authenticates via an exec credential plugin, which is true for EKS but not in general.

When the kubeconfig authenticates by other means (e.g. a client certificate, as used by kops-provisioned nodes), client-go leaves restConfig.ExecProvider as nil. The unconditional access then panics with a nil pointer dereference, crashing the agent on startup before it can connect to the API server.

Guard the rewrite so it only runs when an exec provider is actually present. Exec-based kubeconfigs are unaffected; non-exec kubeconfigs now fall through and use their configured credentials directly.


prasad0896 (commented):
Nice catch, and the writeup nails the root cause. client-go only populates ExecProvider when the kubeconfig authenticates via an exec credential plugin. A standard EKS node always does — the host kubeconfig runs the IAM authenticator as an exec plugin — so this path is safe there. The nil case shows up on kubeconfigs that authenticate some other way, like the client certificate kops-provisioned nodes use, where the unconditional dereference panics the agent on startup. The guard is the right fix and keeps the EKS path untouched.

Before it merges I'd like a test to pin the behavior. This file has no coverage today, which is how the unconditional dereference slipped in, and a refactor could just as easily put it back. Provide() reads the host kubeconfig off disk so it's awkward to test as-is — I'd lift the rewrite into a small helper and cover that:

```
func rewriteExecProvider(restConfig *rest.Config, m chrootMapper) error {
	if restConfig.ExecProvider == nil {
		return nil
	}
	var err error
	restConfig.ExecProvider.Command, restConfig.ExecProvider.Args, err = m.Map(
		restConfig.ExecProvider.Command,
		restConfig.ExecProvider.Args...,
	)
	return err
}
```
Provide() then calls rewriteExecProvider(restConfig, rcp.execMapper), and the test covers both halves:

```
func TestRewriteExecProvider(t *testing.T) {
	t.Run("nil exec provider is left untouched", func(t *testing.T) {
		cfg := &rest.Config{} // client-cert auth, no ExecProvider
		if err := rewriteExecProvider(cfg, chrootMapper{}); err != nil {
			t.Fatalf("unexpected error: %v", err)
		}
		if cfg.ExecProvider != nil {
			t.Fatal("ExecProvider should stay nil")
		}
	})

	t.Run("exec provider is rewritten through the chroot wrapper", func(t *testing.T) {
		cfg := &rest.Config{
			ExecProvider: &clientcmdapi.ExecConfig{
				Command: "aws-iam-authenticator",
				Args:    []string{"token", "-i", "cluster"},
			},
		}
		if err := rewriteExecProvider(cfg, chrootMapper{}); err != nil {
			t.Fatalf("unexpected error: %v", err)
		}
		if filepath.Base(cfg.ExecProvider.Command) != "chroot" {
			t.Fatalf("command not rewritten to chroot wrapper: %q", cfg.ExecProvider.Command)
		}
		if cfg.ExecProvider.Args[1] != "aws-iam-authenticator" {
			t.Fatalf("original command not preserved in args: %v", cfg.ExecProvider.Args)
		}
	})
}
```
The nil case reproduces the panic you fixed; the second makes sure the guard didn't silently disable the rewrite on the EKS path. Happy to approve once that's

grosser:
refactored and added test

prasad0896 (approved):
Thank you for the unit tests. lgtm.

PR #189, opened 2026-06-30
read ca cert from kubectl config

grosser:
**Description of changes**:
atm /etc/kubernetes/pki/ca.crt is hardcoded
user can't override it and the kubectl config that has a field for it is not read

... so let's read it from the kubeclt config

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #191, opened 2026-06-30
fix: resolve kubeconfig client certificate paths against the host mount

grosser:
The pod rest config provider reads the node's kubeconfig to talk to the API server. The cluster CA is now resolved against the host root mount (#189), but the client credentials the kubeconfig references are still opened at their host-absolute paths.

On nodes whose kubeconfig authenticates with a client certificate (e.g. those provisioned by kops) that material is referenced by absolute host path:

    certificate-authority: /srv/kubernetes/api-server-ca-bundle.crt
    client-certificate:    /srv/kubernetes/kubelet.crt
    client-key:            /srv/kubernetes/kubelet.key

The agent reads the host filesystem through a mount, so clientcmd looks for the certificate and key at the wrong location and the provider fails with:

    invalid configuration: [unable to read client-cert /srv/kubernetes/kubelet.crt for kubelet due to open /srv/kubernetes/kubelet.crt: no such file or directory, unable to read client-key /srv/kubernetes/kubelet.key for kubelet due to open /srv/kubernetes/kubelet.key: no such file or directory]

This is not fatal (`main.go` logs it and falls back to the in-cluster ServiceAccount), so it silently loses the host-kubeconfig auth this provider exists to set up.

Resolve the client certificate and key against the host root as well, reusing the prefixing #189 introduced for the CA (extracted as `pathlib.HostPath`). Credentials embedded as `*-data` are self-contained and left untouched, so exec-based EKS kubeconfigs are unaffected.

Rebased onto main now that #189 and #190 have merged: the CA rewriting and the nil-`ExecProvider` guard that this PR originally also carried are upstream, so this is now only the client certificate/key half.

### DIFF
--- a/cmd/eks-node-monitoring-agent/restconfig.go
+++ b/cmd/eks-node-monitoring-agent/restconfig.go
@@ -91,9 +91,9 @@ func rewriteExecProvider(restConfig *rest.Config, m chrootMapper) error {
 }
 
 // hostPathOverrides builds clientcmd overrides that resolve the file paths the
-// host kubeconfig references (CA certificate, client certificate, client key)
-// against the host filesystem mount. Credentials embedded as *-data are
-// self-contained and left untouched.
+// host kubeconfig references (CA certificate, client certificate, client key,
+// token file) against the host filesystem mount. Credentials embedded in the
+// kubeconfig are self-contained and left untouched.
 func (rcp *podRestConfigProvider) hostPathOverrides(kubeconfigPath string) (*clientcmd.ConfigOverrides, error) {
 	hostRoot := config.HostRoot()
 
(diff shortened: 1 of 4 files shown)


grosser:
@prasad0896  rebased this on the merged https://github.com/aws/eks-node-monitoring-agent/pull/189 and re-deployed to check that it still works and is needed 👁️ 🙇

prasad0896:
@grosser sorry for the delay on this one and thank you for rebasing promptly. I have two asks, both in the direction this PR is already going.

HostPath's guard is a string prefix rather than a path-component one, so /hostname/pki/ca.crt
or /host-config/kubelet.key read as already-inside-the-mount and never get prefixed. It came in
with the CA in #189 and behaves the same as before that, so nothing is newly broken - but this
routes the client cert and key through it too, and it's now exported shared code:

```
    func HostPath(hostRoot, path string) string {
        if path == "" || hostRoot == "/" || path == hostRoot || strings.HasPrefix(path, hostRoot+"/") {
            return path
        }
        return filepath.Join(hostRoot, path)
    }
```
And TokenFile is the third host path in AuthInfo, still opened unprefixed. An inline Token takes
precedence over it, so that's the field to guard on:
```
    if len(authInfo.Token) == 0 {
        overrides.AuthInfo.TokenFile = pathlib.HostPath(hostRoot, authInfo.TokenFile)
    }
```

grosser:
`pathlib.HostPath` now compares whole path components, so `/hostname/pki/ca.crt` gets prefixed. It also cleans `hostRoot` first, otherwise `HOST_ROOT=/host/` would stop matching `/host/etc/ca.crt` (the old plain-prefix check happened to handle that).

`TokenFile` is prefixed too, guarded on an empty inline `Token` as you suggested.

Tests added for both: `TestHostPath` in `pkg/pathlib` covers the shared-prefix, host-root-itself and trailing-slash cases, plus two `TestHostPathOverrides` subtests for tokenFile with and without an inline token.

prasad0896 (approved):

PR #199, opened 2026-07-14
feat: make liveness probe timeoutSeconds configurable

dcoppa:
**Description of changes**:

The Kubernetes default of 1 second is not always sufficient on nodes under heavy load, causing spurious liveness probe failures. Add nodeAgent.probeTimeoutSeconds (default 1) to the Helm chart and the add-on configuration schema so it can be raised.

**Testing Done**:

- `helm lint charts/eks-node-monitoring-agent` — passes.
- `helm template` with default values — liveness probe renders `timeoutSeconds: 1`,
  identical to the API-server default that unconfigured pods get today, so existing
  installations see no behavior change:
  ```yaml
  livenessProbe:
    httpGet:
      path: /healthz
      port: 8002
    timeoutSeconds: 1
- helm template --set nodeAgent.probeTimeoutSeconds=5 — renders timeoutSeconds: 5; no other fields in the manifest change.
- Validated charts/configuration.schema.json against sample add-on configurations (python jsonschema):
{"nodeAgent": {"probeTimeoutSeconds": 5}} — accepted
{"nodeAgent": {"probeTimeoutSeconds": 0}} — rejected (below minimum of 1, matching the Kubernetes API constraint)
{"nodeAgent": {"probeTimeoutSeconds": "5"}} — rejected (wrong type)
- Compared the rendered default output against a live pod from a v1.6.6-eksbuild.1 add-on installation to confirm the liveness probe is the only probe in the chart and timeoutSeconds is the only field affected.
- Chart README regenerated with make helm-docs.


prasad0896:
@dcoppa is the motivation for this PR related to something similar to #158 and that is what you are seeing on your nodes?

dcoppa:
> @dcoppa is the motivation for this PR related to something similar to #158 and that is what you are seeing on your nodes?

Yes, something along those lines.

What I'm seeing on my clusters is that the `eks-node-monitoring-agent`
pods occasionally restart because the liveness probe fails.
I initially thought it would be enough to make the probe's
`timeoutSeconds` configurable so it could be increased.
However, if you would prefer the entire probe configuration block to be
configurable, I can work on that instead.

prasad0896:
Got it. I already left some comments about how the probe configuration is not going to solve the problem. We can build on top of my comments in #159 if you are up for it.

Issue #155, opened 2026-05-07
feat(networking-monitor): Add configurable interface exclusion by regexp


TaiSHiNet:
**What would you like to be added**:

Add a configuration option to exclude interfaces from the networking monitor’s `checkInterfaces()` logic by interface-name regexp.

For example:

```yaml
excludedInterfaceNameRegexps:
  - '^ibp[0-9]+s[0-9]+f[0-9]+$'
  - '^ib[0-9]+$'
```

**Suggested behavior:**
* Compile configured regexps at startup.
* If an interface name matches any exclusion regexp, skip InterfaceNotUp / InterfaceNotRunning checks for that interface.
* Log/debug-record that the interface was skipped due to configuration.
* Invalid regexps should fail fast or produce a clear configuration error.

**Why is this needed:**

The networking monitor’s `checkInterfaces()` currently evaluates most interfaces returned by `net.Interfaces()`, excluding some known cases such as Docker-related interfaces, loopback bookkeeping, and interfaces without multicast/broadcast.

On newer accelerated EC2 instance types, nodes may expose additional host-visible interfaces that are not part of Kubernetes node networking. For example, P6 nodes expose Mellanox/NVIDIA InfiniBand IPoIB interfaces such as:
```
ibp115s0f0
ibp116s0f0
```

These interfaces are unrelated to VPC CNI node/pod networking and are not EFA devices. They may legitimately remain down or have no IP address, but NMA currently reports them as InterfaceNotUp / InterfaceNotRunning.

A regexp-based exclusion would let operators suppress known false positives for non-node-networking interfaces without disabling the entire networking monitor.

**Are you currently working around this issue?:**

Yes.

**How are you currently solving this problem?**

Currently the only practical workaround is to modify node boot behavior to prevent the unused IPoIB interfaces from appearing, for example by blacklisting the `ib_ipoib` kernel module. This is broader and riskier than necessary because it changes host networking behavior just to avoid a monitoring false positive.

A targeted NMA-level exclusion would be safer and more explicit.

**Additional context:**

This is intended as an operator-controlled escape hatch. It would preserve the default behavior for existing users while allowing clusters with known non-node-networking interfaces to avoid false-positive node health findings.

**Anything else we should know?**

The specific false positives seen are InterfaceNotUp / InterfaceNotRunning on unused InfiniBand/IPoIB interfaces. The interfaces are present on the host but are not part of Kubernetes node networking, VPC CNI, or EFA device plugin usage.

[...]

PR #202, opened 2026-07-17
feat: add regex based interface filtering

zac-nixon:
**Issue #, if available**:

https://github.com/aws/eks-node-monitoring-agent/issues/155

**Description of changes**:

Adds exactly what the issue is looking for, the ability to turn off interface monitoring using a list regexps.

**Testing Done**:

Validated in my own cluster that I can configure regxps and ignore specific interfaces from monitoring.

### DIFF
--- a/charts/configuration.schema.json
+++ b/charts/configuration.schema.json
@@ -219,6 +219,14 @@
                         "type": "string",
                         "pattern": "^[^/]+/[^/]+$"
                     }
+                },
+                "excludedInterfaceNameRegexps": {
+                    "type": "array",
+                    "description": "List of regular expressions matching interface names that should be excluded from InterfaceNotUp / InterfaceNotRunning checks. Use this to suppress false positives from known non-node-networking interfaces (e.g. Mellanox/NVIDIA IPoIB interfaces such as \"^ibp[0-9]+s[0-9]+f[0-9]+$\"). Defaults to excluding IPoIB interfaces; set to an empty list to disable the default exclusion.",
+                    "default": ["^ibp[0-9]+s[0-9]+f[0-9]+$"],
+                    "items": {
+                      "type": "string"
+                    }
                 }
             }
         },
--- a/pkg/reasons/reasons.yaml
+++ b/pkg/reasons/reasons.yaml
@@ -240,12 +240,20 @@ NetworkingReady:
     Template: 'InterfaceNotRunning'
     DefaultSeverity: 'Fatal'
     Description: >-
-      This interface appears to not be running or there are network issues.
+      This interface appears to not be running or there are network issues. To
+      suppress this for known non-node-networking interfaces (e.g. IPoIB
+      interfaces such as `ibp115s0f0`), set `excludedInterfaceNameRegexps`
+      under `nodeAgent.monitors.networking` in the Helm values or under
+      `monitors.networking` in the config file at `/etc/nma/config.yaml`.
   InterfaceNotUp:
     Template: 'InterfaceNotUp'
     DefaultSeverity: 'Fatal'
     Description: >-
-      This interface appears to not be up or there are network issues.
+      This interface appears to not be up or there are network issues. To
+      suppress this for known non-node-networking interfaces (e.g. IPoIB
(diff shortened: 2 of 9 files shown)


charts/configuration.schema.json
```diff
@@ -1,247 +1,255 @@
...
-                    "$ref": "#/definitions/NodeAgent"
-                },
-                "dcgmAgent": {
-                    "$ref": "#/definitions/DCGMAgent"
-                },
-                "extraObjects": {
-                    "$ref": "#/definitions/ExtraObjects"
-                }
-            }
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
prasad0896: The current diff reindents the entire file, which buries the one addition that matters: the `excludedInterfaceNameRegexps` property under `NetworkingMonitorSettings`. Could you reset the file and add just that property, so the diff is reviewable and blame stays clean?

pkg/reasons/reasons.yaml
```diff
@@ -2,183 +2,183 @@
 # may contain additional metadata that has no functional purpose.
 KernelReady:
   LargeEnvironment:
-    Template: 'LargeEnvironment'
-    DefaultSeverity: 'Warning'
+    Template: "LargeEnvironment"
```
prasad0896: The diff restyles every entry from single to double quotes on top of the two `InterfaceNotUp` / `InterfaceNotRunning` description edits you actually intended. Please revert the whole-file restyle, make just the two description edits, and run make generate so the only delta is those two blocks.

charts/configuration.schema.json
```diff
@@ -1,247 +1,255 @@
...
+          "default": [],
+          "items": {
+            "type": "string",
+            "pattern": "^[^/]+/[^/]+$"
+          }
+        },
+        "excludedInterfaceNameRegexps": {
+          "type": "array",
+          "description": "List of regular expressions matching interface names that should be excluded from InterfaceNotUp / InterfaceNotRunning checks. Use this to suppress false positives from known non-node-networking interfaces (e.g. Mellanox/NVIDIA IPoIB interfaces such as \"^ibp[0-9]+s[0-9]+f[0-9]+$\").",
+          "default": [],
```
prasad0896: The exclusion list defaults to empty today, so every P6 customer still gets Fatal `InterfaceNotUp` / `InterfaceNotRunning` out of the box and has to find this knob before their nodes stop being flagged - which is the toil we set out to remove. Let's default the chart to the known-unusable IPoIB pattern and let customers extend it:
```
nodeAgent:
  monitors:
    networking:
      excludedInterfaceNameRegexps:
        - "^ibp[0-9]+s[0-9]+f[0-9]+$"
```
Keep the pattern tightly anchored so we never mask a real node-networking interface on a future instance family.

prasad0896 (commented):
Nice work - this is the right shape for #155/#157, and reusing the same config pattern as `allowedIPTablesChains` keeps the surface consistent. I do have a few improvement suggestions on the PR before we can merge it in.

charts/eks-node-monitoring-agent/templates/daemonset.yaml
```diff
@@ -46,8 +46,8 @@ spec:
       {{- end }}
       containers:
         - name: {{ .Chart.Name }}
-          image: {{ template "eks-node-monitoring-agent.image" . }}
-          imagePullPolicy: {{ .Values.nodeAgent.image.pullPolicy }}
+          image: 565768096483.dkr.ecr.us-east-1.amazonaws.com/dev:nma
```
prasad0896: We will have to remove this change.

charts/eks-node-monitoring-agent/values.yaml
```diff
@@ -96,7 +96,6 @@ nodeAgent:
   # -- Pod labels applied to the eks-node-monitoring-agent
   podLabels: {}
   # -- Per-monitor configuration keyed by plugin name. See the main README for details.
-  monitors: {}
```
prasad0896, line 99: why are we removing this?

prasad0896 (commented):
new diff looks much more manageable :)

prasad0896 (approved):

Issue #20, opened 2025-09-23
feat: Toggle dcgm daemonset deployment


mhmdhelmy28:
**What would you like to be added**: Toggle dcgm daemonset deployment

**Why is this needed**: We do not host GPU nodes so deploying dcgm is not worth it, I know there's node affinity and it will be scaled to zero if we do not have GPU nodes but we have alerts that check if a daemon set has 0 replicas and we don't want to explicitly exclude dcgm in our alerts if we can just skip deploying them.

How are you currently solving this problem? We're explicitly excluding dcgm from our alerts

ndbaker1:
I understand being able to disable the dcgm ds if you have something running on the host or etc, but i think its a little odd if you're only alarming when `current` replicas == 0, irrespective of the `desired` size?

sidewinder12s:
In our case we already have 1-2 deployments of the DCGM exporter for other reasons.

grosser:
FYI also ran into this, using this as workaround:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

helmCharts:
  - name: eks-node-monitoring-agent
    repo: https://aws.github.io/eks-node-monitoring-agent
    version: 1.6.6
    releaseName: eks-node-monitoring-agent
    namespace: kube-system
    includeCRDs: true
    valuesFile: node-agent-values.yaml

patches:
  # The chart always renders a dcgm-server DaemonSet (no helm value to disable it).
  # We run our own standalone dcgm nv-hostengine, so drop it.
  - target:
      kind: DaemonSet
      name: dcgm-server
    patch: |
      $patch: delete
      apiVersion: apps/v1
      kind: DaemonSet
      metadata:
        name: dcgm-server
  - target:
      kind: DaemonSet
      name: eks-node-monitoring-agent
    patch: |
      # connect to our standalone dcgm nv-hostengine (not the chart's dcgm-server) on port 5567
      - op: add
        path: /spec/template/spec/containers/0/env/-
        value:
          name: DCGM_ADDRESS
          value: localhost:5567

```

PR #205, opened 2026-07-23
feat(charts): support an external DCGM hostengine

dwo:
**Issue #, if available**:

Addresses #20 and the existing-hostengine case in aws/containers-roadmap#2555.

**Description of changes**:

The chart currently deploys the `dcgm-server` DaemonSet, which runs `nv-hostengine` on host port 5555.
The bundled server cannot be disabled, while clusters that already run one hostengine per GPU node must run a second instance.
One example is GPU Operator with standalone DCGM enabled, which exposes its per-node hostengines through the node-local `nvidia-dcgm` Service.

This change adds opt-in values to disable the bundled `dcgm-server`, configure an external DCGM endpoint, and set the node agent DNS policy when the endpoint is a Kubernetes Service.
The external address remains optional so the bundled server can also be disabled when DCGM monitoring is not needed.
Existing defaults preserve the current rendered resources and pod DNS behavior.

**Testing Done**:

Linted and rendered the default, bundled-server-disabled, and external-hostengine configurations with Helm 3.17.2.
Confirmed the default rendering is unchanged and disabling the bundled server works without an external address.
Validated that the agent can connect to GPU Operator standalone DCGM through a node-local Service on EKS.


dwo:
If it's any help for context: we already had `dcgm-exporter` running in our clusters because we are scraping metrics from it to forward to a monitoring solution that we use across multiple cloud providers.

prasad0896:
Thank you for this PR, this is in the right direction to address the two issues you linked. Two things before it merges: rebase off #210, since that moved nv-hostengine into the agent image and changed the dcgm defaults this branch is still sitting on. And expand the docs to spell out the full knob set - how to turn the bundled dcgm agent off
and on (dcgmAgent.enabled), and what address the agent looks for when it's off (dcgmAddress, defaulting to localhost:5555).

grosser:
👁️ @prasad0896 

👍 afaik looks like what we'd want too

prasad0896:
Nice work. I ran both forms on a g4dn node: your own hostengine on the localhost fallback, and a Service hostname
with ClusterFirstWithHostNet. Both work but:

One line to fix before merge. Disabling the bundled server without an address doesn't leave the agent quiet on a GPU node — the nvidia monitor is still registered, so it reports:
```
    AcceleratedHardwareReady=False   reason=DCGMError
    "condition":{"Reason":"DCGMError","Severity":"Fatal","MinOccurrences":0}
```
Fatal, no debounce, about a minute after startup. So the config the doc suggests for "I don't need DCGM" is the one that makes the node eligible for repair. Can you point that case at the per-monitor toggle instead? monitors.nvidia.enabled is already honored and nodeAgent.monitors is in values.yaml:

```
dcgmAgent:
  enabled: false
nodeAgent:
  monitors:
    nvidia:
      enabled: false
```

pjanisze98:
@dwo Bump of this! We are excited to have a solution to this :)

prasad0896 (approved):

prasad0896 (approved):

arsbw2802 (approved):

dwo:
Thanks for seeing this through! I was away on leave 😄

PR #209, opened 2026-07-27
Collect network-policy eBPF maps in the log-collector bundle

yash97:
## Summary

Adds network-policy eBPF map collection to the node monitoring agent's log-collector bundle. On Bottlerocket nodes, support bundles now include the loaded eBPF programs and per-map data.

The agent shells out to `aws-eks-na-cli` to dump the maps. Two builds exist, `aws-eks-na-cli` for IPv4 and `aws-eks-na-cli-v6` for IPv6, and each decodes only its own address family's map keys. The collector must pick the build matching the node before dumping, or the output is garbage.

## How selection works

- A single installed binary (the common case) is used directly.
- When both builds are present, the collector uses a symlink the network policy agent publishes at `/run/aws-network-policy-agent/aws-eks-na-cli`, pointing at the family-correct build. The agent knows the node's family from its own configuration.
- If the family cannot be confirmed, per-map decoding is skipped and the reason is recorded in the bundle. The family-agnostic program and map listing is always collected. The collector never runs an unverified build against a map.

`/usr/bin` is searched before `/opt/cni/bin`: the copy under the CNI directory carries an SELinux label the agent's domain cannot execute, while the `/usr/bin` copy carries one it can.

## Testing

Unit tests cover the selection table and the map-ID parsing. Verified end to end on IPv4 and IPv6 nodes: bundles carry the full listing and per-map data, with no SELinux denials.


e2e/suites/nodediagnostic/ebpf_networkpolicy.go
```diff
@@ -0,0 +1,337 @@
...
+		},
+	}
+
+	return features.New("EbpfMapCollectionWithNetworkPolicy").
+		WithLabel("type", "log-collection").
+		WithSetup("RequireNetworkPolicyAgent", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			if nodeDiagnosticLogBucket == "" {
+				t.Skip("skipping: --nodediagnostic-bucket-name flag was not provided")
+			}
+			if !networkPolicyAgentPresent(ctx, t, cfg) {
```
zac-nixon: How are you validating that this also works in automode?

zac-nixon: Also, is the assumption that an outside test runner will set up NPA?

e2e/suites/nodediagnostic/ebpf_networkpolicy.go
```diff
@@ -0,0 +1,337 @@
...
+const (
+	npServerDeploy   = "np-server"
+	npServerReplicas = 2
+	npPodLabelKey    = "app"
+	npPodLabelVal    = "np-server"
+	awsNodeNamespace = "kube-system"
+	awsNodeDSName    = "aws-node"
+	nodeAgentName    = "aws-eks-nodeagent"
+	// Pinned tag, not :latest, for reproducibility.
+	npServerImage = "public.ecr.aws/nginx/nginx:1.27"
```
zac-nixon: This is going to cause headaches trying to handle CVEs, can I suggest one of the other images, something like https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/0f415d98b80d0178c5c0630ef0d27cf27cb293a3/test/framework/utils/constants.go#L4?

It doesn't look like nginx functionality is actually needed.

yash97: update image.

e2e/suites/nodediagnostic/ebpf_networkpolicy.go
yash97: no test runner will run it, updated logic.

pkg/log_collector/collect/networking_cli.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if err := acc.appendOutput(ebpfDataFile, []byte("*** EBPF loaded data ***\n")); err != nil {
+		return fmt.Errorf("failed to append output to %s: %w", ebpfDataFile, err)
+	}
+	loaded, err := acc.Command(cliPath, "ebpf", "loaded-ebpfdata").CombinedOutput()
+	if err != nil {
+		// Record the failure under the header (e.g. a denied exec on the Auto Mode
+		// managed agent) rather than only in the joined error.
+		return errors.Join(
```
prasad0896: This returns a hard error when the exec fails, which lands in the bundle as log-capture-errors.log (pkg/controllers/nodediagnostic.go collectLogs writes errBuf there and increments subTasksFailed) and makes the whole capture report a failed sub-task. In the e2e, assertLogsValid does `defer t.Fatalf` on that file merely existing, so one denied exec fails the always-run LogCollection suite for the node.

You already made dump-maps best-effort for exactly this reason. Please do the same here: keep the artifact line you're already writing, and return nil. A missing optional diagnostic tool shouldn't mark the bundle as partially failed. modinfo in kernel.go is the existing pattern (CommandOptionsIgnoreFailure).

e2e/suites/nodediagnostic/nodediagnostic.go
```diff
@@ -173,31 +183,126 @@ func LogCollection(awsConfig aws.Config) types.Feature {
...
+
+// assertNoNaCLIDenials fails on an SELinux AVC denial for na-cli in the kernel
+// log. With no dmesg it logs inconclusive rather than passing silently.
+func assertNoNaCLIDenials(t *testing.T, dmesg []byte) {
+	if len(dmesg) == 0 {
+		t.Log("no dmesg collected in bundle; na-cli AVC-denial check is inconclusive")
+		return
+	}
+	for _, line := range strings.Split(string(dmesg), "\n") {
+		if !strings.Contains(line, "avc:") || !strings.Contains(line, "denied") {
```
prasad0896: This is a whole-node kernel-log assertion inside the always-run LogCollection feature, and it matches any line containing `avc:` + `denied` plus either the binary name or `cni_exec_t` - so any denial for any CNI-labelled binary, by any process, fails the shared suite. It will also fire on pre-existing denials that have nothing to do with this change.

Two asks: move it into EbpfMapCollectionWithNetworkPolicy where the collector's exec is actually exercised, and narrow the match to a denial whose target is aws-eks-na-cli AND whose source context is the agent's domain, rather than the bare label.

e2e/suites/nodediagnostic/ebpf_networkpolicy.go
```diff
@@ -0,0 +1,435 @@
...
+
+// nodeOSImage returns the node's OS image for reporting ("unreported OS" if empty).
+func nodeOSImage(node corev1.Node) string {
+	if osImage := node.Status.NodeInfo.OSImage; osImage != "" {
+		return osImage
+	}
+	return "unreported OS"
+}
+
+// enableNetworkPolicy turns on vpc-cni network policy via the addon config so the
```
prasad0896: This mutates the vpc-cni addon on the cluster with ResolveConflicts: OVERWRITE, which rolls the aws-node DaemonSet fleet-wide, and restores best-effort (logging on failure). Three concerns:

  - A failed restore leaves network policy enforcement permanently on for whoever owns the
    cluster.
  - Two concurrent runs race on the snapshot — the case the randomised namespace exists to
    support. Run A snapshots "NP off", B enables, A restores to off underneath B.
  - OVERWRITE can clobber configurationValues another actor manages.

  The in-repo precedent, e2e/suites/addon/configuration_values.go, mutates NMA's *own* addon, passes AddonVersion, doesn't force ResolveConflicts, and t.Fatals on failure. Rolling the dataplane DaemonSet is a much bigger blast radius than that. Could this instead require enforcement to be pre-enabled on the test cluster and skip loudly when it isn't, so the suite reads cluster state rather than mutating shared infrastructure?

e2e/suites/nodediagnostic/ebpf_networkpolicy.go
```diff
@@ -0,0 +1,435 @@
...
+	if nodeagentEnforcesNetworkPolicy(ds) {
+		t.Log("network policy enforcement already enabled on aws-node")
+		return restore
+	}
+	clusterName, err := k8shelper.ExtractClusterName(cfg.KubeContext())
+	if err != nil {
+		t.Logf("could not resolve cluster name to enable network policy: %s", err)
+		return restore
+	}
+	eksClient := eks.NewFromConfig(awsConfig)
```
prasad0896: Missing the stage-aware endpoint override. e2e/suites/addon/configuration_values.go builds its client with awshelper.GetEksEndpoint(stage, ...) as BaseEndpoint; without it these Describe/UpdateAddon calls hit prod endpoints in any non-prod stage, fail, get logged, and the feature silently skips.

e2e/suites/nodediagnostic/ebpf_networkpolicy.go
```diff
@@ -0,0 +1,435 @@
...
+	// Snapshot the addon config first: UpdateAddon replaces the whole
+	// configurationValues document, so we merge enableNetworkPolicy into the
+	// existing config rather than clobbering other keys (WARM_*, custom env, ...),
+	// and restore puts the snapshot back.
+	origConfig, err := vpcCNIConfig(ctx, eksClient, clusterName)
+	if err != nil {
```
prasad0896: Every failure here degrades to a log and then an eventual t.Skip in Assess: no addon, AccessDenied on the newly required eks:DescribeAddon / eks:UpdateAddon, or a 5-minute rollout timeout. So a missing IAM permission or a broken harness produces a permanently skipping test that reads as green, and nobody finds out. Please fail loudly if this e2e is supposed be of value

e2e/setup/e2e.go
```diff
@@ -183,4 +183,8 @@ func TestWrapper(t *testing.T, Testenv env.Environment) {
 		nodediagnostic.LogCollection(awsCfg),
 		nodediagnostic.NodeDestination(),
 	)
+
+	// Forces the network-policy eBPF collector's good path by applying a
+	// NetworkPolicy; skips when the network policy agent is not installed.
+	Testenv.Test(t, nodediagnostic.EbpfMapCollectionWithNetworkPolicy(awsCfg))
```
prasad0896: Any reason this is a separate Testenv.Test call rather than being appended to the existing one above?

prasad0896 (changes requested):

prasad0896 (approved):

PR #210, opened 2026-07-29
feat(dcgm): Run nv-hostengine from the agent image

prasad0896:
**Issue #, if available**:
Slightly related to #82 

**Description of changes**:

The dcgm-server DaemonSet ran nv-hostengine out of eks/observability/dcgm-exporter, which puts an Ubuntu userland and the 77MB dcgm-exporter binary on every GPU node even though the agent only needs the DCGM host engine. Both are scanned surface the agent never executes: the Ubuntu OS packages and the Go dependencies linked into dcgm-exporter.

Ship nv-hostengine, dcgmi, the DCGM modules and nvvs in the eks-node-monitoring-agent image instead, and point dcgm-server at that image. The host engine moves 4.5.2 -> 4.6.1, which also removes an existing skew: the agent's libdcgm client and its go-dcgm bindings are already 4.6.1 while the engine it talked to was 4.5.2.

Rebasing dcgm-exporter onto its distroless variant is not an alternative. That image ships only the exporter binary; it contains no nv-hostengine, so the DaemonSet cannot run at all.

- Dockerfile: install datacenter-gpu-manager-4-core and -proprietary at a pinned DCGM_VERSION, stage the payload with cp -a so the .so symlinks survive (a glob COPY dereferences them and stores every library twice), and copy it as one layer above the Go binaries so agent changes do not invalidate it. The CUDA plugin tiers are left out; the dcgm-exporter image did not carry them either, so diagnostics keep the same scope.
- dcgm-daemonset: exec /usr/bin/nv-hostengine directly instead of wrapping it in a shell with `|| true; sleep infinity`, which kept the container alive around a dead engine. Set the NVIDIA_* variables the exporter image previously carried in its image config; without them the container gets no driver injection and the engine reports zero GPUs.
- values: dcgmAgent.image.tag now defaults to empty, so no dcgm-exporter image is referenced. Setting it, or image.override, pins dcgm-server back to a standalone DCGM image as a break-glass path.
- bump-dcgm-exporter: no-op when no tag is pinned rather than failing. DCGM upgrades now happen through DCGM_VERSION in the Dockerfile.

**Testing Done**:

Verified against the built image: nv-hostengine 4.6.1 starts and serves port 5555, dcgmi connects to it, the Health and Introspection modules load on demand, and the module list matches
Compared it against dcgm-exporter:4.5.2-4.8.1-ubuntu22.04 run on the same host. Default helm rendering changes only this container and contains no dcgm-exporter reference.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,16 +12,38 @@ RUN dnf install -y systemd-devel && \
     dnf clean all
 
 # =============================================================================
-# Stage 2: DCGM builder for GPU monitoring libraries
+# Stage 2: DCGM builder for GPU monitoring libraries and the DCGM host engine
 # =============================================================================
+# NOTE: do not add --platform to this stage. The repository below is selected
+# from `uname -m`, so building on the build platform instead of the target
+# platform would silently place amd64 binaries in the arm64 image.
 FROM public.ecr.aws/amazonlinux/amazonlinux:2023 AS dcgm-builder
 
-# Install DCGM from NVIDIA repository for GPU monitoring support
-# This is optional - the agent works without it on non-GPU nodes
+# DCGM version served to GPU nodes. Pinned for reproducibility: this determines
+# the nv-hostengine version the agent talks to, not just a client library.
+ARG DCGM_VERSION=4.6.1-1
+
+# Install DCGM from the NVIDIA repository for GPU monitoring support.
+# This is optional - the agent works without it on non-GPU nodes.
+#   -core:        libdcgm client, DCGM modules, nv-hostengine, dcgmi, nvvs
+#   -proprietary: plugins/cudaless/libEud.so, for DCGM diagnostics
+# The CUDA plugin tiers (-cudaXX, -proprietary-cudaXX) are deliberately not
+# installed: they add >1GB and the dcgm-exporter image previously used here did
+# not ship them either, so DCGM diagnostics keep the same cudaless-only scope.
 RUN dnf install -y dnf-plugins-core && \
     dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/$(uname -m | sed -e 's/aarch64/sbsa/')/cuda-rhel9.repo && \
-    dnf install -y datacenter-gpu-manager-4-core && \
+    dnf install -y \
+        datacenter-gpu-manager-4-core-1:${DCGM_VERSION} \
+        datacenter-gpu-manager-4-proprietary-1:${DCGM_VERSION} && \
     dnf clean all
+
+# Stage the runtime payload into a single tree. `cp -a` preserves the
+# libfoo.so.4 -> libfoo.so.4.x.y symlinks; copying the globs directly would
+# dereference them and store every library twice.
+RUN mkdir -p /staging/usr/lib64 /staging/usr/bin /staging/usr/libexec && \
+    cp -a /usr/lib64/libdcgm.so* /usr/lib64/libdcgmmodule*.so* /staging/usr/lib64/ && \
+    cp -a /usr/bin/nv-hostengine /usr/bin/dcgmi /staging/usr/bin/ && \
+    cp -a /usr/libexec/datacenter-gpu-manager-4 /staging/usr/libexec/
 
 # =============================================================================
 # Stage 3: Go builder to compile the application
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -12,17 +12,39 @@ RUN dnf install -y systemd-devel && \
...
-    dnf install -y datacenter-gpu-manager-4-core && \
+    dnf install -y \
+        datacenter-gpu-manager-4-core-1:${DCGM_VERSION} \
+        datacenter-gpu-manager-4-proprietary-1:${DCGM_VERSION} && \
     dnf clean all
 
+# Stage the runtime payload into a single tree. `cp -a` preserves the
+# libfoo.so.4 -> libfoo.so.4.x.y symlinks; copying the globs directly would
+# dereference them and store every library twice.
+RUN mkdir -p /staging/usr/lib64 /staging/usr/bin /staging/usr/libexec && \
```
fletcherw, line 43: why the 2 step copy? what breaks if you directly copy or install, similar to what happens with systemd-builder below?

e2e/setup/manifests/agent.tpl.yaml
```diff
@@ -541,10 +541,17 @@ spec:
       priorityClassName: system-node-critical
       containers:
         - name: eks-node-monitoring-agent-dcgm
-          image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/observability/dcgm-exporter:4.5.2-4.8.1-ubuntu22.04
+          image: {{ .Image }}
           imagePullPolicy: IfNotPresent
-          command: ["/bin/sh"]
-          args: ["-c", "nv-hostengine -n -b ALL || true; sleep infinity"]
+          command: ["/usr/bin/nv-hostengine"]
```
fletcherw, line 546: you've dropped the /bin/sh wrapper, why is that?

Dockerfile
```diff
@@ -12,17 +12,39 @@ RUN dnf install -y systemd-devel && \
...
+#   -proprietary: plugins/cudaless/libEud.so, for DCGM diagnostics
+# The CUDA plugin tiers (-cudaXX, -proprietary-cudaXX) are deliberately not
+# installed: they add >1GB and the dcgm-exporter image previously used here did
+# not ship them either, so DCGM diagnostics keep the same cudaless-only scope.
 RUN dnf install -y dnf-plugins-core && \
     dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/$(uname -m | sed -e 's/aarch64/sbsa/')/cuda-rhel9.repo && \
-    dnf install -y datacenter-gpu-manager-4-core && \
+    dnf install -y \
+        datacenter-gpu-manager-4-core-1:${DCGM_VERSION} \
+        datacenter-gpu-manager-4-proprietary-1:${DCGM_VERSION} && \
```
fletcherw, line 37: proprietary here is what provides nv-hostengine? why are we copying dcgmi as well?

Dockerfile
prasad0896: Nothing breaks - COPY dereferences symlinks matched by a glob, so each of DCGM's nine `libfoo.so.4` -> `libfoo.so.4.6.1` links would get stored as a full second copy of the library (~22 MB raw / ~8 MB compressed of duplication, on every node whether it has a GPU or not). Staging with cp -a preserves the symlinks and collapses it into one layer; I built it both ways and the image is 163 MB vs 193 MB. systemd-builder has the same behaviour but only ~4.5 MB of it, so I left it alone to keep this diff scoped. Installing directly in the runtime stage isn't possible since eks-distro-minimal-base-glibc has no dnf/rpm.

e2e/setup/manifests/agent.tpl.yaml
prasad0896: The wrapper hid failures. With nv-hostengine -n -b ALL || true; sleep infinity, an engine that failed to start or exited left the container Running on sleep infinity, so Kubernetes saw a healthy pod while the agent reported DCGMError and nothing restarted. `-n `already keeps the engine in the foreground, so the shell and the sleep were never needed, and running the binary directly makes it PID 1 so it receives SIGTERM and a failure surfaces as a container restart. The DaemonSet only schedules onto GPU instance types, so there is no node where the engine exiting is expected.

Dockerfile
prasad0896: `-core` is what provides nv-hostengine, along with dcgmi, nvvs, libdcgm and the DCGM modules. `-proprietary` only adds `plugins/cudaless/libEud.so` for DCGM diagnostics; I included it because the dcgm-exporter image we are replacing installed both packages (its dpkg metadata lists datacenter-gpu-manager-4-core and datacenter-gpu-manager-4-proprietary), so diagnostics keep the same scope. dcgmi is there because the e2e nvidia suite execs dcgmi inside the dcgm-server pod, which worked with the old image, and it also gives us `kubectl exec ds/dcgm-server -- dcgmi diag` for debugging.

Dockerfile
fletcherw: makes sense, nice work catching the improvement.

fletcherw (approved):
approving pending passing CI

prasad0896:
CI passed, failed in tear down.

PR #220, opened 2026-08-12
fix: bound collector execution so a blocked collector cannot stop diagnostics

### LINKED ISSUE #219: Console diagnostics and NodeDiagnostic capture stop permanently when any collector blocks (no timeout on exec.Command)
Repo: `aws/eks-node-monitoring-agent` · observed on **v1.6.6**

Line references below are pinned to the `v1.6.6` tag. Still present on `main` as of
`b7dea82` — the producer loop is unchanged and all five subprocess collectors still use
`exec.Command` (`systemdStatus`, `journalctl`, `cpuUsage`, `diskUsage`, `dmesg`).

## Summary

`diagnosticLogger.Start` runs its collectors sequentially in a single goroutine, and none
of them can time out. If any collector blocks indefinitely, the loop never reaches its
ticker and **console diagnostics stop for the remaining life of the node** — silently, with
no error, no log line, and no partial output for the sections that would have followed.

A hung NFS/EFS mount is enough to trigger it: `df -T` blocks in uninterruptible sleep, so
the `disk` collector never returns.

**The same defect affects `NodeDiagnostic` log capture**, which is a separate code path
(`collectLogs` → `pkg/log_collector/collect/*`) that converges on the same unbounded
`exec.Command` in `pkg/osext`. There the impact is larger: the bundle hangs with
`CaptureState` stuck in `Running`, so the primary support-bundle mechanism is unusable on
exactly the nodes that need it. Details, code path and an observed instance are in
[this comment](https://github.com/aws/eks-node-monitoring-agent/issues/219#issuecomment-5251531149).

## Observed impact

On an EKS Auto Mode node (Bottlerocket), one unresponsive EFS mount stopped console
diagnostics **11 minutes after boot**. Nothing was emitted for the following ~12 days.
`ps` on day 12 still showed the original blocked process:

```
root  424046  0.0  0.0  5780  2816 ?  D  Jul30  0:15  df -T
```

The serial console shows the cycle stopping mid-way, exactly at a producer boundary — the
last two sections emitted are `cpu` and `memory`, and `disk` (the next producer) plus all
seven sections after it never appear again:

```
NMA::LOG|2026-07-30T08:53:45Z|cpu
NMA::LOG|2026-07-30T08:53:45Z|memory
   <nothing further, ever>
```

Kernel output continued reaching `/dev/console` for days afterwards, so the console itself
was fine — only the agent's writer goroutine was stuck.

Because the collectors are ordered, a block in an early producer also suppresses every
later one. `disk` is third of ten, so the node lost `interfaces`, `ipamd`, `apiserver`,
`dmesg`, `systemd`, `kubelet` and `containerd` as well. Those are precisely the sections
needed to debug a sick node, and this failure mode is most likely exactly when the node is
sick.

## Root cause

[`pkg/diagnostic/logger.go#L94-L113`](https://github.com/aws/eks-node-monitoring-agent/blob/v1.6.6/pkg/diagnostic/logger.go#L94-L113)
— producers are called inline, so `producer.fn()` blocking blocks the whole loop:

```go
for {
    for _, producer := range producers {
        ...
        data := tailx(producer.fn(), sectionBytes)   // no timeout, no isolation
        ...
    }
    select {
    case <-ticker.C:                                 // never reached
    case <-ctx.Done():
    }
}
```

Every subprocess collector uses `exec.Command` rather than `exec.CommandContext`, so none
of them is bounded:

| Collector | Line | Call |
|---|---|---|
| `cpuUsage` | [L132-133](https://github.com/aws/eks-node-monitoring-agent/blob/v1.6.6/pkg/diagnostic/logger.go#L132-L133) | `exec.Command("ps", "aux")` |
| `diskUsage` | [L140-141](https://github.com/aws/eks-node-monitoring-agent/blob/v1.6.6/pkg/diagnostic/logger.go#L140-L141) | `exec.Command("df", "-T")` |
| `dmesg` | [L177-178](https://github.com/aws/eks-node-monitoring-agent/blob/v1.6.6/pkg/diagnostic/logger.go#L177-L178) | `exec.Command("dmesg")` |
| `systemdStatus` | [L115-117](https://github.com/aws/eks-node-monitoring-agent/blob/v1.6.6/pkg/diagnostic/logger.go#L115-L117) | `exec.Command("systemctl", ...)` |
| `journalctl` | [L124-125](https://github.com/aws/eks-node-monitoring-agent/blob/v1.6.6/pkg/diagnostic/logger.go#L124-L125) | `exec.Command("journalctl", ...)` |

`df -T` is the most exposed, since it stats every mount and a single unresponsive one
blocks it forever. But the problem is general, not specific to `df`:

- `ps aux` walks `/proc` and can block on a task stuck in the kernel.
- `dmesg`, `systemctl`, `journalctl` can all stall under I/O pressure or a wedged journal.
- [`ipamd()` at L185-186](https://github.com/aws/eks-node-monitoring-agent/blob/v1.6.6/pkg/diagnostic/logger.go#L185-L186)
  reads a file with no bound; if that path ever lands on unresponsive storage it blocks too.
  (Now `readFileTail` on `main` via #196 — bounded in size, but still an unbounded blocking
  read, so the point stands.)

`testAPIServerEndpoint` is the only collector that is already bounded (its `http.Client`
has `Timeout: 5 * time.Second`).

## Suggested fix

1. **Bound every subprocess** — use `exec.CommandContext` with a per-collector timeout
   comfortably below `LogInterval` (default 5 min,
   [L44](https://github.com/aws/eks-node-monitoring-agent/blob/v1.6.6/pkg/diagnostic/logger.go#L44)).
   Note that `CommandContext` alone does not rescue a process wedged in uninterruptible
   sleep — the kill will not land — so the timeout must also release the *loop*, not just
   signal the child.
2. **Isolate collectors from the loop** — run each in its own goroutine and `select` on
   completion vs. a deadline, so a permanently stuck collector degrades to one missing
   section instead of stopping the agent. This is the part that actually guarantees
   forward progress.
3. **Emit the failure** — on timeout, write the section header with an explicit
   `collector timed out after Ns` body, matching the existing convention of returning
   `failed to call ... due to: %s` as section content. Today a hung collector is
   indistinguishable from the agent being dead.
4. Optionally reorder or parallelise so one slow collector cannot suppress the others.

Happy to open a PR if the maintainers agree on the approach.

youwalther65:
**Issue #, if available**: Fixes #219

**Description of changes**:

Collectors in both affected paths now run under an execution bound. `exec.CommandContext` alone is not enough: a process in uninterruptible sleep never receives the kill that cancellation sends, so `cmd.Wait` cannot return. Each bound therefore releases its caller and abandons what it cannot interrupt.

- **`pkg/osext`** — adds `CommandContext` for chrooted commands, plus `Output`/`CombinedOutput` bounded by a timeout, returning `ErrTimeout`. `Command` is unchanged, so the monitors are unaffected.
- **`pkg/diagnostic`** — producers take a context and each runs isolated under a 30s bound, so a stuck collector costs its own section (`collector "disk" timed out after 30s`) instead of every later one. A producer whose previous run is still outstanding is skipped rather than started again, so a command that cannot be killed is not respawned every cycle for the life of the node.
- **`cmd/.../main.go`** — the deferred flush on exit called `Start`, which emitted one cycle and then blocked on its ticker forever, so a crashing agent never exited. It now calls a new `RunOnce` under a 30s bound.
- **`pkg/log_collector` + `pkg/controllers`** — the `Accessor` carries the collection context and a 60s `CommandTimeout`, replacing the direct `exec` calls in `CommandOutput` and in the six collectors that ran commands themselves. A stuck command now returns an error, which is what the existing per-collector isolation needs to do its job: the capture continues and names the failure in `log-capture-errors.log` and `subTasksFailed`. `LogCaptureTimeout` (20m) backstops a collector that blocks outside a subprocess, so `CaptureState` completes as a failure naming the timeout instead of staying `Running`.

Timeouts are set well above what these commands legitimately need, so hitting one means the command is stuck rather than slow.

**Testing Done**:

New unit tests cover each bound: a blocked collector no longer stops the console cycle and is not restarted while outstanding, a healthy collector still runs every cycle, shutdown is reported as shutdown rather than as a timeout, `CommandOutput` fails with `ErrTimeout` instead of hanging (and `IgnoreFailure` still tolerates it), and a blocked capture completes as `Failure` naming the timeout.

`make test` and `go test ./pkg/... -race -count=1` are green on Linux/amd64 with Go 1.26.4, including `helm lint`.

Not covered by unit tests: no test creates a genuinely unkillable process, since that needs a real unresponsive NFS/EFS mount. What the tests prove is that the caller is released at the deadline; the uninterruptible-sleep case rests on abandoning the goroutine rather than waiting on it.

### DIFF
--- a/cmd/eks-node-monitoring-agent/main.go
+++ b/cmd/eks-node-monitoring-agent/main.go
@@ -65,6 +65,10 @@ var (
 
 const (
 	envNodeName = "MY_NODE_NAME"
+
+	// consoleFlushTimeout bounds the final diagnostic cycle written on exit, so
+	// that a collector blocked on host I/O cannot hold up the process teardown.
+	consoleFlushTimeout = 30 * time.Second
 )
 
 func init() {
@@ -84,8 +88,13 @@ func main() {
 			f, _ := openDevConsole()
 			defer f.Close()
 			// ensures that at least one run of the diagnostic is always
-			// completed and written to console before exiting.
-			diagnostic.NewDiagnosticLogger(f, diagnostic.Settings{}).Start(context.Background())
+			// completed and written to console before exiting. RunOnce emits a
+			// single cycle and returns; Start would block on its ticker and keep
+			// a crashing agent from ever exiting. The bound keeps a collector
+			// that is wedged on host I/O from delaying the exit indefinitely.
+			flushCtx, cancelFlush := context.WithTimeout(context.Background(), consoleFlushTimeout)
+			defer cancelFlush()
+			diagnostic.NewDiagnosticLogger(f, diagnostic.Settings{}).RunOnce(flushCtx)
 
 			// increase visibility on errors/panics propagated from main.
 			if r := recover(); r != nil {
--- a/pkg/controllers/nodediagnostic.go
+++ b/pkg/controllers/nodediagnostic.go
@@ -57,6 +57,16 @@ type nodeDiagnosticController struct {
 	// logCaptureTimeout overrides LogCaptureTimeout. Zero means the default; it
 	// is only set in tests, which cannot wait out the real bound.
 	logCaptureTimeout time.Duration
+
+	// captureRunning reports whether a previous log collection has not returned
+	// yet. A collector blocked on unresponsive host I/O never returns, so
+	// without this every later NodeDiagnostic would start a fresh collection and
+	// leak both a goroutine and the temp directory that collection is writing
+	// into, for the life of the node.
+	captureRunning atomic.Bool
+	// captureStartedAt is the unix nano timestamp of the outstanding collection,
+	// and is only meaningful while captureRunning is true.
+	captureStartedAt atomic.Int64
 }
 
 func NewNodeDiagnosticController(kubeClient client.Client, nodeName string, runtimeContext *config.RuntimeContext) *nodeDiagnosticController {
@@ -579,11 +589,41 @@ const LogCaptureTimeout = 20 * time.Minute
 // collection is deliberately not archived — it writes into the temp directory
 // that collectLogs removes on return, and archiving a directory with a live
 // writer in it would race the archiver against a growing file.
+//
+// Abandoning is only safe because of captureRunning. An abandoned collection
+// still holds its goroutine and its temp directory: collectLogs removes that
+// directory with a deferred call which does not run until the blocked read
+// returns, and on a wedged mount it never does. While a collection is
+// outstanding this reports the block instead of starting another one, so a
+// permanently wedged collector costs one goroutine and one temp directory
+// rather than a fresh pair for every NodeDiagnostic.
 func (c *nodeDiagnosticController) collectLogsBounded(ctx context.Context, categories []v1alpha1.LogCategory) (io.Reader, int, error) {
 	timeout := c.logCaptureTimeout
 	if timeout <= 0 {
 		timeout = LogCaptureTimeout
 	}
+
+	if !c.captureRunning.CompareAndSwap(false, true) {
+		// zero means the winning caller has claimed the slot but has not stored its
+		// start time yet, so there is no duration to report. Reporting one anyway
+		// would measure from the epoch.
+		if startedAt := c.captureStartedAt.Load(); startedAt == 0 {
+			return nil, 0, fmt.Errorf(
+				"log collection from a previous NodeDiagnostic has not returned and is still blocked on host I/O; " +
+					"not starting another until it does")
+		} else {
+			blocked := time.Since(time.Unix(0, startedAt)).Truncate(time.Second)
+			return nil, 0, fmt.Errorf(
+				"log collection from a previous NodeDiagnostic has not returned after %s and is still blocked on host I/O; "+
+					"not starting another until it does", blocked)
+		}
+	}
+	// stored by the caller that claimed the slot, so the timestamp belongs to the
+	// outstanding collection. Storing it before the claim would let every refused
+	// caller overwrite it with the time of its own attempt, which reads back as
+	// zero elapsed; the refusal path above handles the not-yet-stored case instead.
+	c.captureStartedAt.Store(time.Now().UnixNano())
+
 	// not named "collect": that would shadow the log collector package.
 	collectFn := c.collectFunc
 	if collectFn == nil {
@@ -602,6 +642,9 @@ func (c *nodeDiagnosticController) collectLogsBounded(ctx context.Context, categ
 	done := make(chan result, 1)
 	go func() {
 		archive, issueCount, err := collectFn(collectCtx, categories)
+		// released before publishing, so that a caller which receives the result
+		// always observes the collection as free again.
+		c.captureRunning.Store(false)
 		done <- result{archive, issueCount, err}
 	}()
 
--- a/pkg/controllers/nodediagnostic.go
+++ b/pkg/controllers/nodediagnostic.go
@@ -49,6 +49,14 @@ type nodeDiagnosticController struct {
 	// captureFunc is the function called to execute the packet capture. It defaults to
 	// captureAndUploadPackets but can be overridden in tests to avoid running tcpdump.
 	captureFunc func(ctx context.Context, nd *v1alpha1.NodeDiagnostic, captureID string) ([]error, error)
+
+	// collectFunc is the function called to collect logs. It defaults to
+	// collectLogs but can be overridden in tests to avoid reading host state.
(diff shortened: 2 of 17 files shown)


prasad0896:
Nice call abandoning the goroutine instead of waiting on it — the producer.running guard is
what makes that safe.

collectLogsBounded doesn't have that guard, so a NodeDiagnostic against a wedged mount leaves
a goroutine and its capture dir behind on every attempt. CopyFile/CopyDir are also still
unbounded — os.Open and io.Copy, no context, on host paths.

Can you bound those two and give collectLogsBounded the same outstanding-run guard collect() uses?

Nit: `startedAt` is stored after the CompareAndSwap, so a skip in that window reports from zero.

youwalther65:
One more thing worth flagging, which I'd rather raise than have you find.

  The exit-path snapshot ([main.go#L86-L97](https://github.com/aws/eks-node-monitoring-agent/blob/c91702e/cmd/eks-node-monitoring-agent/main.go#L86-L97), the `defer` around `RunOnce`) constructs its own NewDiagnosticLogger, so it does not share the running/startedAt flags with the periodic loop. On a node where a collector is already wedged, that snapshot will start a fresh one rather than skipping it — so "one stuck process, not one per cycle" holds for the periodic loop, plus one per agent exit.

  I've left it that way deliberately. The exit snapshot is the last output a crashing agent produces, and suppressing a section because it was stuck earlier could hide the thing being diagnosed. It's bounded by the 30s flush timeout, so it cannot delay shutdown. Note the defer fires on any return, not only a panic, so this is once per exit including ordinary restarts.

  Happy to make the two share producer state if you'd prefer the guarantee to be unconditional — it would mean threading one logger through to the exit path rather than building a second.

  Two smaller questions while you're looking:
  
  1. CopyFile on cancellation. The context is checked before os.Create, so a copy cancelled before it starts leaves nothing behind. A copy cancelled mid-stream necessarily leaves a truncated file. Would you rather that partial file were removed, or is a short file in the bundle more useful than an absent one? Note `753f59d` leaves it in place and its test now asserts that, so this is a decision to make rather than an open gap.
  2. The bundle guard vs MaxConcurrentReconciles: 1. This might look redundant. It isn't: that option serialises reconcile calls, and an abandoned collection outlives the call that started it - collectLogsBounded returns at its bound while the collection goroutine stays blocked in the read. So the next reconcile is correctly serialised and still finds a live collection. Arguably the guard restores the invariant the existing comment on that option already claims ("mutating actions are fully completed before processing the next item"), which an abandoned goroutine breaks. Wanted to confirm that's the reading you intended.

Edit: link repointed to `c91702e` after the branch was rebased onto `fe2129f` and the review commits squashed; typo fixed.

youwalther65:
@prasad0896 Thanks — all three addressed in `753f59d`.

**`collectLogsBounded` guard.** It now carries the same `running`/`startedAt` pair as `collect()`, released in the goroutine before publishing so a completed collection frees the slot immediately. Worth noting the leak was slightly worse than a goroutine: `collectLogs` removes its temp directory with a deferred call, which does not run until the blocked read returns, so each abandoned attempt also stranded a partial capture tree. The refusal names how long the outstanding collection has been blocked, so it reaches `NodeDiagnostic` status rather than leaving `CaptureState` in `Running` with no explanation.

**`CopyFile`/`CopyDir`.** Both take the accessor context now. One limitation I would rather state than have you find: the bound is **cooperative**. `os.Open` and a read already in the kernel are ordinary blocking syscalls, so a source on a wedged mount can still block inside one of them. What the context buys is stopping at the next read boundary instead of copying a whole file, and stopping the descent instead of reading every remaining directory. Bounding a copy that is *already* stuck is still the caller abandoning the collection goroutine — which is why the guard above matters for these too. The check sits before `MkdirAll` and `ReadDir` as well as per entry, since `ReadDir` is itself a call that blocks on a wedged mount, and creating the destination first left empty directories behind that read as genuinely empty log directories in the bundle.

**`startedAt` nit.** I tried moving the store before the compare-and-swap first, and it is worse: the store is unconditional, so every refused caller overwrites the outstanding run's start time immediately before reading it back. A collector blocked for three seconds then reports `after 0s`, which loses the only information the message carries — whether a node has been wedged for thirty seconds or six hours. So the store stays after the claim, and instead the skip path treats a zero timestamp as "claimed but not yet stamped" and omits the duration:

```
collector "disk" has not returned and is skipped until it does
collector "disk" has not returned after 5m0s and is skipped until it does
```

Both windows are now correct, with a test for each.

`go test -race` green on all three packages, on Linux and macOS. The branch is rebased onto `fe2129f` and the review changes squashed into one commit, so earlier SHAs in this thread are stale.

prasad0896 (approved):
lgtm

PR #222, opened 2026-08-12
fix: close the apiserver response body and reuse one client

### LINKED ISSUE #221: Console diagnostics: apiserver check never closes the response body and builds a new http.Transport per call
## Summary

[`testAPIServerEndpoint`](https://github.com/aws/eks-node-monitoring-agent/blob/main/pkg/diagnostic/logger.go#L227-L248) does not close the HTTP response body, and constructs a fresh `http.Client` and `http.Transport` on every call. The `apiserver` section runs once per `LogInterval` (5 minutes by default), and console diagnostics are always enabled on EKS Auto Mode nodes.

Pinned to `main` at `b7dea82`, [pkg/diagnostic/logger.go#L227-L248](https://github.com/aws/eks-node-monitoring-agent/blob/main/pkg/diagnostic/logger.go#L227-L248):

```go
client := &http.Client{
	Timeout:   5 * time.Second,
	Transport: &http.Transport{ /* ... */ },   // new Transport per call
}
r, err := client.Get(url)
if err != nil { /* ... */ }
body, err := io.ReadAll(r.Body)                // no defer r.Body.Close()
```

## Consequences

- **No connection reuse.** A new `Transport` per call has its own pool, so every cycle pays a fresh TCP connect and TLS handshake to the API server, forever.
- **Client-side cleanup is delegated entirely to the peer.** The body is read to EOF, which returns the connection to the discarded transport's idle pool, where its read-loop goroutine keeps it alive. Because the hand-constructed `Transport` leaves `IdleConnTimeout` at zero — no limit, unlike `http.DefaultTransport`'s 90s — nothing on the client will ever reap it. Today it is the API server's own 90s `IdleTimeout` ([secure_serving.go#L171](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/secure_serving.go#L171)) that closes it, so in practice at most one such connection is outstanding between 5-minute cycles.
- **The error path is worse.** If `io.ReadAll` fails part-way, the body is neither drained nor closed, so the connection is not returned to the pool at all and waits on the peer to drop it.

So this is not an unbounded fd leak against kube-apiserver, but it does mean correctness rests on a peer behaviour the agent does not control — any intermediary or endpoint configuration without an idle timeout changes the outcome — and it repeats a TLS handshake every five minutes for no reason.

## Suggested fix

- `defer r.Body.Close()` after the error check, per `net/http`'s documented contract.
- Build the client once (package level, or in `NewDiagnosticLogger`) and reuse it, so connections are pooled and the handshake is not repeated.

## Notes

Separate from #196, which fixed file and journald handle leaks in this same logger but did not touch this function.

Happy to open a PR.

youwalther65:
**Issue #, if available**: Fixes #221

**Description of changes**:

The apiserver console-diagnostics check built an `http.Client` and `http.Transport` per call and never closed the response body. Each check therefore had its own connection pool, so every cycle paid a fresh TCP connect and TLS handshake, and the connection was left for the peer to close — a hand-built transport leaves `IdleConnTimeout` at zero, meaning no limit, so nothing client-side would ever reclaim it. Today the API server's own 90s idle timeout is what does.

- The client is built once at package level and reused.
- `IdleConnTimeout` is set to 90s, matching `http.DefaultTransport` and the API server's own idle timeout, so the client no longer depends on the peer to reap idle connections.
- The body is closed with a `defer`, so it is released whether or not the read succeeds.

The check runs once per `LogInterval` (5 minutes by default) for the life of a node with console diagnostics enabled, which is always the case on EKS Auto Mode.

**Testing Done**:

Added a test that serves the check from an `httptest` TLS server and counts new connections via `ConnState`: three sequential checks now share one connection, where before they opened three. Verified it fails without the production change (`expected: 1, actual: 3`), so it covers both the client reuse and the missing `Close` — reuse is impossible unless the body is released. A second test covers the no-response path, where there is no body to close.

`go test ./pkg/diagnostic/ -race` is green.

### DIFF
--- a/pkg/diagnostic/logger.go
+++ b/pkg/diagnostic/logger.go
@@ -224,22 +224,33 @@ func readFileTail(path string, n int64) []byte {
 	return buf
 }
 
-func testAPIServerEndpoint(host string) []byte {
-	client := &http.Client{
-		Timeout: 5 * time.Second,
-		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{
-				// Skipping TLS verification is ok as we are just validating the
-				// connectivity to the API server
-				InsecureSkipVerify: true,
-			},
+// apiServerClient is built once and reused for every check. A client and
+// transport per call would give each check its own connection pool, so every
+// cycle would pay a fresh TCP connect and TLS handshake for the life of the node.
+var apiServerClient = &http.Client{
+	Timeout: 5 * time.Second,
+	Transport: &http.Transport{
+		TLSClientConfig: &tls.Config{
+			// Skipping TLS verification is ok as we are just validating the
+			// connectivity to the API server
+			InsecureSkipVerify: true,
 		},
-	}
+		// A hand-built transport leaves this at zero, which means no limit, so
+		// idle connections would only ever be reclaimed by the peer closing them.
+		// Matches http.DefaultTransport and the API server's own idle timeout.
+		IdleConnTimeout: 90 * time.Second,
+	},
+}
+
+func testAPIServerEndpoint(host string) []byte {
 	url := fmt.Sprintf("%s/livez?verbose", host)
-	r, err := client.Get(url)
+	r, err := apiServerClient.Get(url)
 	if err != nil {
 		return []byte(fmt.Sprintf("failed to make request endpoint due to: %s\n", err))
 	}
+	// the body must be closed for the connection to be released, whether or not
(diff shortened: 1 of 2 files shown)


prasad0896:
The `Body.Close` is the real fix here for #221 . Thanks for the change

prasad0896 (approved):

Issue #223, opened 2026-08-12
bug(kernel-monitor): AppCrash false positives from intentional int3 traps; empty process name for real segfaults


grosser:
## Summary

The kernel monitor's `appCrash` matcher treats *every* `traps:` line in `dmesg` as an application crash. This includes `trap int3`, which is a **deliberate** userspace abort (`__builtin_trap()` / Chromium's `base::ImmediateCrash()` / a failed `CHECK`), not a fault and not a node health problem.

On nodes running browser-based CI (Selenium grid / headless Chrome test pods), each browser session teardown emits one `trap int3` line, so the agent produces a continuous stream of `KernelReady` / `AppCrash` warning events.

While investigating this I also found the process name is reported as an empty string for the one case that *is* a genuine crash — a real segfault. Details in "Bug 2" below.

## Environment

* Agent built from current `main` (`monitors/kernel/monitor.go` unchanged from `main` at time of writing)
* Ubuntu 24.04.4 LTS, kernel `6.8.0-1062-aws`, containerd 2.2.2, amd64
* Kubernetes v1.36

## Reproduction / observed behaviour

`dmesg -T` on an affected node, one line roughly every 3 seconds:

```
[Wed Aug 12 19:48:53 2026] traps: chrome[1356759] trap int3 ip:5f9ad8850f04 sp:7fff57cb87c0 error:0 in chrome[5f9ad6161000+c48d000]
[Wed Aug 12 19:48:57 2026] traps: chrome[1361870] trap int3 ip:587223a07f04 sp:7ffe4686e870 error:0 in chrome[587221318000+c48d000]
[Wed Aug 12 19:49:00 2026] traps: chrome[1366886] trap int3 ip:5ef8c4c3df04 sp:7ffcaad69f50 error:0 in chrome[5ef8c254e000+c48d000]
```

Resulting events:

```
$ kubectl get events -A --field-selector involvedObject.kind=Node
NAMESPACE   LAST SEEN   TYPE      REASON         OBJECT                                MESSAGE
default     11m         Warning   KernelReady    node/ip-10-x-x-x.ec2.internal         AppCrash: Process "chrome" on the node has crashed
```

Event counts over a ~20h window on one cluster, 10 affected nodes, ~5,500 events total:

```
ip-10-x-x-120   469   2026-08-12T00:08:49Z -> 2026-08-12T19:48:47Z
[...]

PR #225, opened 2026-08-13
fix(kernel): do not report intentional int3 traps as AppCrash

grosser:
Fixes the `int3` half of #223. Minimal on purpose: it does not touch the capture-group bug or the missing deduplication also described there.

## Problem

`appCrash` accepts any `traps:` line:

```go
appCrash = regexp.MustCompile(strings.Join([]string{
	`traps:\s*(.*?)\[`,
	`\s(.*?)\[\d+]: segfault at.*`,
}, "|"))
```

But `trap int3` is a breakpoint trap applications raise **deliberately** — `__builtin_trap()`, or Chromium's `IMMEDIATE_CRASH()` on a failed `CHECK`. It is not a fault and carries no information about node health.

On nodes running browser-based CI this means a continuous stream of `KernelReady` / `AppCrash` warnings. Measured on one cluster over ~20h: **~5,500 events across 10 nodes**, worst single node 1,511, every one of them headless Chrome ending a test session:

```
[Wed Aug 12 19:48:53 2026] traps: chrome[1356759] trap int3 ip:5f9ad8850f04 sp:7fff57cb87c0 error:0 in chrome[5f9ad6161000+c48d000]
[Wed Aug 12 19:48:57 2026] traps: chrome[1361870] trap int3 ip:587223a07f04 sp:7ffe4686e870 error:0 in chrome[587221318000+c48d000]
```

All 711 traps on the node I examined hit the identical offset in the Chrome binary (`ip - base = 0x26eff04`, 711/711) — a single deterministic code path, once per browser session, which is what you would expect from a fixed `CHECK` rather than from anything going wrong on the host. I attributed the processes to a CI pod by sampling `/proc/<pid>/cgroup`, and the traps stop entirely when no test pods are scheduled.

## Fix

Skip `int3` specifically. Go's `regexp` has no negative lookahead, so this is a separate pattern rather than an inline exclusion:

```go
// int3 is a breakpoint trap that applications raise deliberately, for example via
// __builtin_trap() or Chromium's IMMEDIATE_CRASH(), so it does not indicate a node problem
intentionalTrap = regexp.MustCompile(`traps:.* trap int3\b`)
```

Genuine faults are unaffected — verified directly:

```
appCrash=true intentional=true   | traps: chrome[1450999] trap int3 ip:5c6363ff7f04 ...
appCrash=true intentional=false  | traps: myapp[42] trap divide error ip:400abc ...
appCrash=true intentional=false  | traps: myapp[42] general protection fault ip:400abc ...
appCrash=true intentional=true   | traps: weird[7] trap int3
```

The `\b` rather than a trailing space covers `int3` at end of line.

I deliberately did **not** use an allowlist of fault types (`divide error`, `general protection fault`, ...), even though #223 floats that as the tidier option. An allowlist silently drops any trap type not enumerated, and `trap invalid opcode` is genuinely ambiguous — `ud2` is emitted both by compiler-inserted trap paths and intentionally by some language runtimes. Denying the one type known to be intentional keeps current behaviour for everything else. Happy to switch to an allowlist if maintainers prefer it, including or excluding `invalid opcode` as you see fit.

## Testing

Added `TestKernelMonitor/AppCrashIgnoresIntentionalTrap`, using a real dmesg line, asserting no notification is emitted.

Full package passes on Linux (`golang:1.26` container, since the package needs `unix.Adjtimex`/`STA_UNSYNC` and will not build on darwin):

```
ok  	github.com/aws/eks-node-monitoring-agent/monitors/kernel	0.096s
```

I also confirmed the test is not vacuous — reverting just the one-line condition makes it fail:

```
--- FAIL: TestKernelMonitor/AppCrashIgnoresIntentionalTrap
    Error: Not equal: expected: 0, actual: 1
```

`go vet` is clean for the package. `GOOS=linux go build ./...` fails only inside the NVIDIA go-dcgm cgo dependency, which is a pre-existing cross-compilation limitation unrelated to this change.

## Note on scope

This stops the agent from misreporting the traps. It does not stop Chrome from raising them — that is a workload-side problem I am fixing separately. The two remaining items in #223 (segfault reporting `Process ""` because `matches[1]` is the wrong capture group, and the lack of per-process deduplication) are untouched here and I am happy to send them as follow-ups.

### DIFF
--- a/monitors/kernel/monitor.go
+++ b/monitors/kernel/monitor.go
@@ -89,6 +89,9 @@ var (
 		// dmesg timestamp prefix
 		`\s([^\s\[]+)\[\d+]: segfault at.*`,
 	}, "|"))
+	// int3 is a breakpoint trap that applications raise deliberately, for example via
+	// __builtin_trap() or Chromium's IMMEDIATE_CRASH(), so it does not indicate a node problem
+	intentionalTrap = regexp.MustCompile(`traps:.* trap int3\b`)
 )
 var (
 	appBlocked        = regexp.MustCompile(`task (.*?):\d+ blocked for more than`)
@@ -111,7 +114,7 @@ func (k *KernelMonitor) handleDmesg(line string) error {
 				Message("A kernel bug was detected and reported by the Linux kernel").
 				Build(),
 		)
-	} else if matches := appCrash.FindStringSubmatch(line); matches != nil {
+	} else if matches := appCrash.FindStringSubmatch(line); matches != nil && !intentionalTrap.MatchString(line) {
 		processName := appCrashProcessName(matches)
 		return k.manager.Notify(context.Background(),
 			reasons.AppCrash.
--- a/monitors/kernel/monitor.go
+++ b/monitors/kernel/monitor.go
@@ -83,15 +83,16 @@ var (
 	)
 	appCrash = regexp.MustCompile(strings.Join([]string{
 		// each top level group is expected to have one sub group capture for the
-		// process name, use appCrashProcessName to read whichever one matched
-		`traps:\s*(.*?)\[`,
+		// process name, use appCrashProcessName to read whichever one matched.
+		// the fault names are a bounded allowlist from arch/x86/kernel/traps.c. int3
+		// is left out because it is a breakpoint applications raise deliberately (for
+		// example Chromium's IMMEDIATE_CRASH()), not a node problem. divide error is
+		// printed with a "trap " prefix, the rest are not.
+		`traps:\s*(.*?)\[\d+] (?:trap )?(?:divide error|general protection fault|stack segment|segment not present|invalid TSS|alignment check|bounds)`,
 		// anchored on the bracketed pid so the capture cannot grow left into the
 		// dmesg timestamp prefix
 		`\s([^\s\[]+)\[\d+]: segfault at.*`,
 	}, "|"))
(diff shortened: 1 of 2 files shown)


prasad0896 (dismissed):

prasad0896:
Thanks for pointing this out and the PR. One nit, non-blocking: the trap names come from a fixed table in
arch/x86/kernel/traps.c, so an allowlist of the faults we care about is bounded where a denylist
isn't. __builtin_trap() compiles to ud2 and prints "trap invalid opcode", so the case your comment
cites still fires. Swapping the traps: arm for this drops the need for the int3 skip:
```
    `traps:\s*(.*?)\[\d+] (?:trap )?(?:general protection fault|divide error|stack segment|alignment check|segment not present|invalid TSS)`
```
I left invalid opcode out -- an illegal instruction is a wrong-binary problem, not a node one --
but keep it if you disagree. Happy either way.

Heads up that this and #226 rewrite the same else-if line, so you will have to fix this anyways :)

grosser:
lets get 226 merged then and I'll come back after

grosser:
updated, will run the fork monday and see if it still fixes the issue I saw

grosser:
yeah works, all the false-positives are gone :)
image

prasad0896 (approved):

prasad0896:
Thanks for the rebase and update to allowlist. Much more cleaner and controlled :)

PR #226, opened 2026-08-13
fix(kernel): report the real process name for AppCrash segfaults

grosser:
Fixes the capture-group half of #223.

## Problem

`appCrash` joins two patterns with `|`, and each carries its own capture group:

```go
appCrash = regexp.MustCompile(strings.Join([]string{
	// each top level group is expected to have one sub group capture for the
	// process name
	`traps:\s*(.*?)\[`,
	`\s(.*?)\[\d+]: segfault at.*`,
}, "|"))
```

The consumer only ever reads group 1:

```go
processName := matches[1]
```

Alternation numbering does not work that way — group 1 belongs to the first alternation only. A **segfault** matches the second alternation, which captures into group 2, so group 1 is empty and the event reads:

```
AppCrash: Process "" on the node has crashed
```

The comment above the regex suggests group 1 was *intended* to be the process name for every branch, which is presumably how this went unnoticed.

There is a second bug in the same pattern: `\s(.*?)` is unanchored on the left, so the capture grows past the process name into the dmesg timestamp prefix. Using the exact line already in the test table:

```
input:  [   32.298491][  T896] kexec[896]: segfault at 0 ip 0000000000000000 sp 00007ffeaf0ff420 error 14 in dash[561ac3c57000+4000]
groups: ["   32.298491][  T896] kexec[896]: segfault at 0 ..." "" "  32.298491][  T896] kexec"]
```

So even `matches[2]` would have been wrong — `"  32.298491][  T896] kexec"` instead of `"kexec"`. Both bugs had to be fixed to get a correct name out of a segfault.

Net effect today: `AppCrash` is the agent's signal for a genuine application fault, and for the one line format that unambiguously *is* a fault, the event names no process at all.

## Fix

1. Anchor the segfault capture on the bracketed pid so it cannot grow leftward:

   ```go
   `\s([^\s\[]+)\[\d+]: segfault at.*`
   ```

2. Read whichever group the matching alternation actually set:

   ```go
   // appCrash alternates between patterns that each carry their own capture group, so only the
   // group belonging to the alternation that matched is set and the rest are empty.
   func appCrashProcessName(matches []string) string {
   	for _, processName := range matches[1:] {
   		if processName != "" {
   			return processName
   		}
   	}
   	return ""
   }
   ```

I scanned by group index rather than hardcoding `matches[2]` deliberately: the original bug is exactly what happens when an alternation is added and the index is not revisited, so this stays correct if a third pattern shows up.

## Testing

The table test asserted only reason and severity, which is why it passed while rendering `Process ""`. Added an optional `message` field and assertions for both alternations — the existing segfault line, plus a `traps:` line to keep branch 1 covered:

```go
{"[   32.298491][  T896] kexec[896]: segfault at 0 ...", "AppCrash", ..., `Process "kexec" on the node has crashed`},
{"[Wed Aug 12 19:49:58 2026] traps: chrome[1450999] trap divide error ...", "AppCrash", ..., `Process "chrome" on the node has crashed`},
```

Verified each fix is independently necessary by reverting one at a time:

```
# with matches[1] restored
expected: "Process \"kexec\" on the node has crashed"
actual  : "Process \"\" on the node has crashed"

# with the unanchored capture restored
expected: "Process \"kexec\" on the node has crashed"
actual  : "Process \"  32.298491][  T896] kexec\" on the node has crashed"
```

Full package passes on Linux (`golang:1.26` container — the package needs `unix.Adjtimex`/`STA_UNSYNC` and will not build on darwin):

```
ok  	github.com/aws/eks-node-monitoring-agent/monitors/kernel	0.108s
```

`go vet` is clean for the package.

I used a `traps: ... trap divide error` line rather than `int3` for the new branch-1 case, so this PR stays orthogonal to #225 and the two do not conflict.

## Not included

The remaining item in #223 is the lack of per-process deduplication (a crash-looping process emits hundreds of near-identical events). Happy to send that separately.


prasad0896 (approved):

grosser:
workflows need another approval

PR #229, opened 2026-08-18
feat(manager): Support condition recovery in node exporter

prasad0896:
**Issue #, if available**:

**Description of changes**:

A monitor can report a fatal reason but has no way to withdraw one. Once a managed NodeCondition goes False it stays False for the life of the process, even after the underlying problem clears, and a condition failing for two separate reasons cannot be partially cleared at all.

Add a `Resolved` flag to `monitor.Condition` and a `Resolve` method to the `Exporter` interface. The node exporter tracks unresolved fatal reasons per condition type in arrival order, so resolving one reason rebuilds the condition message from the reasons that remain, and resolving the last one restores the configured ready state with a Normal event for auditability. The manager routes resolved conditions past `MinOccurrences` gating, resets that reason's debounce counter, and clears the fatal condition gauge on full recovery. Message aggregation and `LastTransitionTime` semantics are unchanged, and nothing emits a resolved condition yet.

**Testing Done**:

New exporter tests cover resolving one reason out of several, resolving the last reason, and resolving a reason that was never reported. The existing flapping regression test passes unchanged. `make test`, `hack/check-generate.sh`, `gofmt -l` and `GOOS=linux CGO_ENABLED=1 go build ./...` are green on this branch.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -133,6 +133,22 @@ func (m *MonitorManager) exportCondition(ctx context.Context, monitorName string
 	}
 	logger = logger.WithValues("conditionType", conditionType)
 
+	// Resolved conditions clear state instead of reporting it: reset the
+	// occurrence debounce for the reason and route to the exporter's resolve
+	// path, bypassing MinOccurrences gating.
+	if condition.Resolved {
+		m.conditionCountMap[condition.Reason] = 0
+		recovered, err := m.exporter.Resolve(ctx, condition, conditionType)
+		if err != nil {
+			return err
+		}
+		if recovered {
+			conditionTypeGauge.WithLabelValues(string(conditionType)).Set(0)
+		}
+		logger.Info("resolved condition", "recovered", recovered)
+		return nil
+	}
+
 	// Skip requests for conditions that have not met their minimum occurrences
 	if m.conditionCountMap[condition.Reason] < condition.MinOccurrences {
 		logger.Info("condition has not met MinOccurrences", "occurrences", m.conditionCountMap[condition.Reason])
--- a/pkg/manager/node_exporter.go
+++ b/pkg/manager/node_exporter.go
@@ -48,6 +48,8 @@ func NewNodeExporter(
 		recorder:               recorder,
 		managedConditions:      initializeManagedConditions(managedConditionConfigs),
 		managedConditionsDirty: true,
+		conditionConfigs:       managedConditionConfigs,
+		fatalEntries:           make(map[corev1.NodeConditionType][]fatalEntry),
 	}
 }
 
@@ -92,6 +94,20 @@ type nodeExporter struct {
 	managedConditions      map[corev1.NodeConditionType]corev1.NodeCondition
 	managedConditionsDirty bool
 	managedConditionsLock  sync.Mutex
+
+	// conditionConfigs holds the ready-state configuration used to restore a
+	// condition once all of its fatal reasons have been resolved.
+	conditionConfigs map[corev1.NodeConditionType]NodeConditionConfig
+	// fatalEntries tracks the unresolved fatal reasons per condition type in
+	// arrival order, so that resolving one reason does not clear a condition
+	// that is still failing for another.
+	fatalEntries map[corev1.NodeConditionType][]fatalEntry
+}
+
+// fatalEntry is one unresolved fatal reason and its latest message.
+type fatalEntry struct {
+	reason  string
+	message string
 }
 
 // Info records an event for the specified condition.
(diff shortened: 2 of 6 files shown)


pkg/manager/node_exporter.go
```diff
@@ -108,36 +124,120 @@ func (e *nodeExporter) Warning(ctx context.Context, c monitor.Condition, conditi
...
+		entries = append(entries, fatalEntry{reason: monitorCondition.Reason, message: monitorCondition.Message})
+	}
+	e.fatalEntries[conditionType] = entries
+
+	e.rebuildFatalCondition(conditionType)
+	return nil
+}
+
+// Resolve clears a previously reported fatal reason for the condition type.
+// When the last reason is cleared, the condition is restored to its healthy
```
fletcherw, line 152: Are any conditions/reasons ever shared between multiple monitors? Could one monitor resolving clear another's valid fatal condition if they were?

pkg/manager/node_exporter.go
```diff
@@ -108,36 +124,120 @@ func (e *nodeExporter) Warning(ctx context.Context, c monitor.Condition, conditi
...
+	updated := false
+	for i := range entries {
+		if entries[i].reason == monitorCondition.Reason {
+			entries[i].message = monitorCondition.Message
+			updated = true
+			break
+		}
+	}
+	if !updated {
+		entries = append(entries, fatalEntry{reason: monitorCondition.Reason, message: monitorCondition.Message})
```
fletcherw, line 143: are these messages dynamic? could a single monitor spamming the same fatal condition with slightly different messages exhaust memory?

pkg/manager/node_exporter.go
```diff
@@ -108,36 +124,120 @@ func (e *nodeExporter) Warning(ctx context.Context, c monitor.Condition, conditi
...
+		if !duplicate && entry.message != "" {
+			messages = append(messages, entry.message)
+		}
+	}
+
 	newCondition := corev1.NodeCondition{
 		Type:               conditionType,
-		Reason:             monitorCondition.Reason,
-		Message:            monitorCondition.Message,
+		Reason:             entries[len(entries)-1].reason,
```
fletcherw, line 229: newest fatal reason is what shows on the NodeCondition? Why not oldest?

pkg/manager/node_exporter.go
prasad0896: They can't be. Every monitor maps 1:1 to a condition type, and the only two that name the same one are mutually exclusive at registration: nvidia is skipped unless `AcceleratedHardware()` reports nvidia, neuron unless it reports neuron, and that field holds a single value. A node has one accelerator type, so only one of the two is ever registered. Reason identifiers are unique in `reasons.yaml`, so there's no path for one monitor to resolve another's entry.

pkg/manager/node_exporter.go
prasad0896: No, because entries are keyed by reason rather than by message. `Fatal()` finds the existing entry for that reason and replaces its message in place, so a monitor re-firing the same reason with different text overwrites instead of appending. The slice only grows with distinct reasons, and those come from the fixed taxonomy in `reasons.yaml`. Event emission has a separate bound arriving in #234, where each component gets a fixed per-node budget.

pkg/manager/node_exporter.go
prasad0896: `NodeCondition.Reason` is a single string, so one has to win. Nothing is lost: every unresolved reason's message is aggregated into the condition `Message`, so the full set is visible there. We surface the most recently added unresolved reason because that's the change an operator is looking for first when they read the node.

pkg/manager/node_exporter.go
fletcherw: ah I misread that it was keyed only on reason, thanks.

pkg/manager/node_exporter.go
fletcherw: but oldest condition is probably more likely to be root cause, while newest is more likely to be just a symptom. I think both choices are reasonable but I'd urge you to consider it more.

fletcherw (dismissed):

pkg/manager/node_exporter.go
Pavani-Panakanti: Here when a reason thats already tracked is reported again https://github.com/aws/eks-node-monitoring-agent/blob/55a8fc1cec10b6866ee762f9f3b40d6ccf7c03b1/pkg/manager/node_exporter.go#L136 we are just updating the message but not changing its position in fatalentries list. That means reason is always taken from last and refires are not considered as recent/newest. If there is some reason that needs no action thats pinned and some reason that needs action that keeps firing it will essentially be masked ?

pkg/manager/manager.go
```diff
@@ -133,6 +133,22 @@ func (m *MonitorManager) exportCondition(ctx context.Context, monitorName string
...
+	// occurrence debounce for the reason and route to the exporter's resolve
+	// path, bypassing MinOccurrences gating.
+	if condition.Resolved {
+		m.conditionCountMap[condition.Reason] = 0
+		recovered, err := m.exporter.Resolve(ctx, condition, conditionType)
+		if err != nil {
+			return err
+		}
+		if recovered {
+			conditionTypeGauge.WithLabelValues(string(conditionType)).Set(0)
```
Pavani-Panakanti, line 146: When resolved this will be set to 0 but in line 128 we are bumping counditioncount 
```
conditionCount.WithLabelValues(string(condition.Severity), condition.Reason).Add(1)
```
We should move that to after this resolved block so both counter and guage stay in sync and do not contradict

PR #230, opened 2026-08-18
feat(probe): Add probe spec types and startup validation

prasad0896:
**Issue #, if available**:

**Description of changes**:

Detection for the AWS-owned agents that ship on the node is written by hand per agent, so onboarding one means new Go code inside a monitor rather than a declaration. Nothing rejects a bad health-check definition either — a typo in a reason name would simply never emit.

Introduce `api/probe` with a declarative Spec covering the checks, the reason emitted on failure, the interval and the thresholds, plus `pkg/probe` validation that runs at startup. Specs are pure data and serialize to JSON and YAML through `metav1.Duration`, so registration can move from binary-baked defaults onto the configuration surface later without reshaping the contract. Validation resolves the failure reason against the generated lookup, rejects parameterized reason templates a probe cannot render, and enforces the address and path rules for each transport.

**Testing Done**:

Validation tests cover each rejection path, and the Spec is round-tripped through both JSON and YAML to confirm the serialized shape. `make test`, `hack/check-generate.sh`, `gofmt -l` and `GOOS=linux CGO_ENABLED=1 go build ./...` are green on this branch.

### DIFF
--- a/api/probe/probe.go
+++ b/api/probe/probe.go
@@ -1,0 +1,92 @@
+// Package probe defines the declarative contract for subagent health probes.
+//
+// A probe declares how the node monitoring agent observes the health of one
+// AWS-owned subagent (e.g. IPAMD, the Network Policy Agent) through the
+// agent's own health surface. Probes are pure data: the framework in
+// pkg/probe interprets a Spec and emits conditions through the parent
+// monitor, so onboarding an agent is a spec addition rather than new
+// detection code.
+package probe
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/aws/eks-node-monitoring-agent/api/monitor"
+)
+
+// TransportKind identifies how a Check reaches its target agent.
+type TransportKind string
+
+const (
+	// TransportHTTPLoopback performs an HTTP GET against a local-only
+	// address. A 2xx response is healthy; any other response, including a
+	// refused connection, is unhealthy.
+	TransportHTTPLoopback TransportKind = "http-loopback"
+	// TransportSystemdDBus queries the ActiveState of a systemd unit over
+	// D-Bus. Liveness only: "active" is healthy, anything else is unhealthy.
+	// A failure to reach D-Bus itself is unknown, not unhealthy.
+	TransportSystemdDBus TransportKind = "systemd-dbus"
+)
+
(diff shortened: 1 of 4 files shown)


api/probe/probe.go
```diff
@@ -0,0 +1,92 @@
...
+	// Interval is the time between check rounds.
+	Interval metav1.Duration `json:"interval"`
+
+	// FailureThreshold is the number of consecutive unhealthy results
+	// required before the failure condition is emitted. The counter resets
+	// on any healthy result. Must be at least 1.
+	FailureThreshold int `json:"failureThreshold"`
+
+	// StartupGracePeriod suppresses failure emission for this duration after
+	// the runner starts, tolerating agents that come up after the monitoring
```
fletcherw, line 62: this seems flaky. setting absolute time delays in config will never have a right answer. How could you express "this check shouldn't run until X happens", e.g. some systemd unit is started?

for healthchecker, we retry checks until a unit is initially healthy, could something similar work here?

prasad0896: As discussed offline, for now we can proceed with this implementation. Although it is arbitrary and slightly flaky, it is currently a slightly better approach than hand waving this in code for various agents based on implementation. 

We can always improve this down the line as we onboard new agents.

fletcherw (dismissed):

Pavani-Panakanti:
Can we address copilot comments as well ? Thanks

api/probe/probe.go
```diff
@@ -0,0 +1,92 @@
...
+// Checks defines the individual checks that make up a probe. Liveness is
+// required; readiness and diagnostics are optional because not every
+// transport can express them (systemd ActiveState has no notion of
+// readiness).
+type Checks struct {
+	// Liveness answers "is the agent running".
+	Liveness Check `json:"liveness"`
+	// Readiness answers "is the agent performing its function". The depth of
+	// the readiness answer is owned by the agent, not by this contract.
+	Readiness *Check `json:"readiness,omitempty"`
```
Pavani-Panakanti, line 76: Severity is per-spec, so liveness and readiness share it. If liveness needs to be Fatal, readiness is Fatal too, which means no check in a spec can ever produce a Warning event, it emits only condition. Want to confirm thats the intended behavior i.e. the probe framework is for repair-worthy checks only and anything event-shaped stays in a one-off handler or just don't add it?

PR #231, opened 2026-08-18
feat(probe): Add http-loopback and systemd-dbus transports

prasad0896:
**Issue #, if available**:

**Description of changes**:

A probe has to reach its target, and the two shapes present on the node today are an HTTP endpoint on loopback and a systemd unit's `ActiveState` over D-Bus. The networking monitor already queries `ActiveState` in two places with the logic duplicated between them.

Add a `Transport` interface returning a three-valued outcome. `Healthy` and `Unhealthy` are both answers about the agent, and a refused connection to an agent's own health endpoint counts as `Unhealthy`. `Unknown` is deliberately distinct: the check could not execute for reasons on the monitoring side, such as an unreachable D-Bus socket, and those results must not count toward a failure threshold, because being unable to ask the question is not evidence about the agent. The http-loopback transport treats 2xx as healthy and captures a bounded portion of a non-2xx response body into the result detail. The systemd-dbus transport generalizes the duplicated `ActiveState` query and adds best-effort `SubState` and `Result` enrichment when a unit is not active.

**Testing Done**:

Transport tests cover healthy, non-2xx with body detail, connection refused, and cancelled-context paths for HTTP, and active, inactive-with-enrichment, connection error, property error, and unexpected-type paths for D-Bus via a fake connection. `make test`, `hack/check-generate.sh`, `gofmt -l` and `GOOS=linux CGO_ENABLED=1 go build ./...` are green on this branch.

### DIFF
--- a/pkg/probe/transport_dbus.go
+++ b/pkg/probe/transport_dbus.go
@@ -1,0 +1,67 @@
+package probe
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/coreos/go-systemd/v22/dbus"
+
+	"github.com/aws/eks-node-monitoring-agent/api/probe"
+)
+
+// dbusConn is the subset of the go-systemd dbus connection used by the
+// transport, extracted so tests can substitute a fake.
+type dbusConn interface {
+	GetUnitPropertyContext(ctx context.Context, unit string, propertyName string) (*dbus.Property, error)
+	Close()
+}
+
+// SystemdDBusTransport executes liveness checks by querying a systemd unit's
+// ActiveState over D-Bus. It generalizes the ActiveState query used by the
+// networking monitor's IPAMD and NPA handlers. Failure to reach D-Bus itself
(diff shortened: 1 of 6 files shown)


pkg/probe/transport_dbus.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (t *SystemdDBusTransport) Do(ctx context.Context, check probe.Check) Result {
+	conn, err := t.newConn(ctx)
+	if err != nil {
+		return Result{Outcome: OutcomeUnknown, Detail: fmt.Sprintf("connecting to D-Bus: %v", err)}
+	}
+	defer conn.Close()
+
+	property, err := conn.GetUnitPropertyContext(ctx, check.Address, "ActiveState")
+	if err != nil {
```
Pavani-Panakanti, line 46: can we filter any of these errors to unhealthy like unit not found or not loaded or do all errors fall into unknown here

PR #232, opened 2026-08-18
feat(probe): Add the probe runner and its failure semantics

prasad0896:
**Issue #, if available**:

**Description of changes**:

With the contract and the transports in place, something has to schedule checks and decide when a probe failure becomes a node condition. Detection latency needs to be predictable, an agent that starts after the monitoring agent must not trip a failure during boot, and a flapping agent must not be able to evade repair.

The runner executes one spec on its interval using the shared jittered ticker. A consecutive-failure threshold resets on any healthy result, making detection latency threshold times interval. A startup grace period suppresses emission while agents come up. `Unknown` results move neither counter. Liveness runs first and readiness only when liveness is healthy, with failures attributed by check name in the condition message.

Two behaviors keep events cheap and repair reachable. A failure streak that recovers before reaching the threshold emits exactly one Warning summarizing the completed episode, since the probed agent cannot record its own brief outage; streaks that begin inside the grace period are boot noise and emit nothing. `RecoveryThreshold` sets how many consecutive healthy results are required before a fired failure resolves, because every False to True transition resets the repair wait clock and resolving on the first healthy tick would let a flapping agent avoid repair indefinitely. Zero defaults to one and preserves the simpler behavior.

**Testing Done**:

Runner tests are table-driven on a single cycle with an injected clock and ticker, so none of them depend on wall-clock timing. They cover emission after consecutive failures, success resetting the counter, resolution after recovery, recovery hysteresis, `Unknown` counting toward neither threshold, grace-period suppression, episode summaries below the threshold, summary suppression for streaks beginning in grace, readiness failures, and exit on context cancellation. `make test`, `hack/check-generate.sh`, `gofmt -l` and `GOOS=linux CGO_ENABLED=1 go build ./...` are green on this branch.

### DIFF
--- a/pkg/probe/runner.go
+++ b/pkg/probe/runner.go
@@ -16,15 +16,20 @@ import (
 // Runner executes one probe spec on its interval and emits conditions
 // through the parent monitor's manager. It owns the probe's failure
 // semantics: a startup grace period, a consecutive-failure threshold that
-// resets on success, and a recovery signal once a previously failing probe
-// is healthy again. Unknown results (checks that could not execute for
-// monitoring-side reasons) never count toward the threshold.
+// resets on success, a consecutive-healthy recovery threshold that damps
+// condition flapping, and a one-shot Warning summarizing failure episodes
+// that recover before reaching the failure threshold. Unknown results
+// (checks that could not execute for monitoring-side reasons) never count
+// toward either threshold.
 type Runner struct {
 	spec       probe.Spec
 	transports map[probe.TransportKind]Transport
 	manager    monitor.Manager
 	log        logr.Logger
 	severity   monitor.Severity
+	// recoveryThreshold is spec.RecoveryThreshold with the zero value
+	// defaulted to 1 (resolve on first healthy result).
+	recoveryThreshold int
 
 	// now and newTicker are injectable for tests.
 	now       func() time.Time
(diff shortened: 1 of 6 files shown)


pkg/probe/runner.go
```diff
@@ -0,0 +1,209 @@
...
+			// agent cannot record its own brief outage, so emit exactly one
+			// Warning summarizing the completed episode — unless the streak
+			// began inside the startup grace period, which is boot noise.
+			startedInGrace := r.streakStartedInGrace
+			r.streakStartedInGrace = false
+			if startedInGrace {
+				r.log.Info("probe episode began during startup grace, suppressing summary", "failures", streak)
+				return nil
+			}
+			duration := time.Duration(streak) * r.spec.Interval.Duration
```
Pavani-Panakanti, line 131: should this be (streak-1)*interval if we are checking duration from first to last ?

pkg/probe/runner.go
```diff
@@ -0,0 +1,209 @@
...
+		if grace := r.spec.StartupGracePeriod.Duration; r.now().Sub(r.startedAt) < grace {
+			r.log.Info("probe failing within startup grace period, suppressing", "detail", result.Detail, "consecutiveFailures", r.consecutiveFailures)
+			return nil
+		}
+		if r.consecutiveFailures < r.spec.FailureThreshold {
+			r.log.Info("probe failing below threshold", "detail", result.Detail, "consecutiveFailures", r.consecutiveFailures)
+			return nil
+		}
+		r.failureEmitted = true
+		return r.manager.Notify(ctx, monitor.Condition{
```
Pavani-Panakanti, line 172: Once the threshold is crossed this re-notifies on every tick, so at a 30s interval a persistently failing probe drives a node status patch every 30s for as long as the failure lasts. The exporter marks conditions dirty on every Fatal without comparing against the current condition, and it refreshes LastHeartbeatTime each time

Could the runner skip the notify when the detail is unchanged and failureEmitted is already true, and let the exporter's 5-minute heartbeat own the timestamp refresh? According to design timestamp refresh is supposed to happen every 5 min ?

PR #233, opened 2026-08-18
feat(networking): Migrate Auto Mode IPAMD liveness to probe framework

prasad0896:
**Issue #, if available**:

**Description of changes**:

Auto Mode IPAMD liveness is an inline D-Bus `ActiveState` query inside `handleIPAMD`, debounced by a fifteen-minute consistency window that exists to absorb noise from the DaemonSet path's process scan. On Auto Mode the query is deterministic and does not need that window, and recovery never restores `NetworkingReady` because nothing withdraws the reason.

Register the equivalent probe instead: systemd-dbus liveness of `ipamd.service` on a thirty-second interval, three consecutive failures, five-minute startup grace to tolerate the service coming up after the monitoring agent. Detection tightens from roughly five to fifteen minutes down to ninety seconds. The probe reuses the `IPAMDNotRunning` reason and inherits Fatal from the taxonomy, so customer auto-repair configuration keyed on that reason is unaffected, and recovery now restores the condition. What stays in `handleIPAMD` on Auto Mode is the checkpoint and CRI cross-check, which reads a file IPAMD wrote rather than asking the agent about itself. The DaemonSet path is untouched.

**Testing Done**:

A test asserts the Auto Mode spec shape (subsystem, reason, transport, address) and that the DaemonSet path registers no probe. The existing `IPAMDNotRunning` monitor test passes unchanged, covering the DaemonSet process-scan path. `make test`, `hack/check-generate.sh`, `gofmt -l` and `GOOS=linux CGO_ENABLED=1 go build ./...` are green on this branch.

### DIFF
--- a/monitors/networking/probes.go
+++ b/monitors/networking/probes.go
@@ -1,0 +1,41 @@
+package networking
+
+import (
+	"slices"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiprobe "github.com/aws/eks-node-monitoring-agent/api/probe"
+	"github.com/aws/eks-node-monitoring-agent/pkg/config"
+)
+
+// probeSpecs returns the subagent health probes supervised by the networking
+// monitor for the current delivery path. Probes replace only agent-mediated
+// liveness detection; checkpoint cross-checks and configuration validation
+// remain handler-based.
+func (m *NetworkingMonitor) probeSpecs() []apiprobe.Spec {
+	if !slices.Contains(m.runtimeContext.Tags(), config.EKSAuto) {
+		// On the DaemonSet path IPAMD runs inside the VPC CNI pod and is
+		// observed via the existing process scan in handleIPAMD; its gRPC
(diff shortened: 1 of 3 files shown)


monitors/networking/probes.go
```diff
@@ -0,0 +1,41 @@
...
+					Address:   "ipamd.service",
+				},
+			},
+			ReasonOnFail:     "IPAMDNotRunning",
+			Interval:         metav1.Duration{Duration: 30 * time.Second},
+			FailureThreshold: 3,
+			// Tolerate ipamd.service coming up after the monitoring agent
+			// during boot; systemd ordering does not sequence the two.
+			StartupGracePeriod: metav1.Duration{Duration: 5 * time.Minute},
+		},
```
Pavani-Panakanti, line 39: RecoveryThreshold will be default which is 1. That means we will miss cases where ipamd is having issues but had one healthy check in between. Can we evaluate raising RecoveryThreshold to higher value ?
We will still miss some flapping cases but better than 1

PR #234, opened 2026-08-18
feat(manager): Add per-component event budgets

prasad0896:
**Issue #, if available**:

**Description of changes**:

Every event this agent records on a node shares a single client-go spam-filter budget. The filter keys on the event source, the involved object and the event type, not on the reason, so all monitors and probes on a node compete for one 25-event burst refilled at one token per five minutes. A single component in a failure loop can starve every other component's events for hours, and the drops are silent — no error reaches the caller and no metric moves.

Make the component the unit of the budget. Every reason in `reasons.yaml` now declares its owning component, so ownership lives in the same ledger as the taxonomy, and the generator fails when declared plus reserved components exceed the platform capacity. Onboarding capacity becomes a deliberate ledger change rather than something the next agent discovers through silent drops. The exporter shapes emission ahead of the recorder with one token bucket per component and event type as the guaranteed budget, plus a per-type node ceiling derived from that quota as a bug detector. A flooding component starves only itself, and every drop is counted per component on new `nma_events_*` metrics.

Enforcement sits in the exporter rather than in the correlator's spam key on purpose. The correlator runs aggregation before the spam filter and rebuilds aggregated events without annotations, so a component-keyed spam filter would be defeated by the correlator itself and its drops would stay invisible. The custom broadcaster is still installed, but only to key aggregation on the component and to size the spam filter as a backstop above the exporter's own shaping. Event reason and message are byte-identical to today; the component travels as an `eks.amazonaws.com/component` annotation.

**Testing Done**:

An isolation test drives one component to its burst limit and confirms a second component's event still delivers immediately, and that the flooding component's other event type is unaffected. A second test drives every declared component to its burst and asserts the node ceiling never rejects, which is the invariant that makes a ceiling rejection a bug rather than capacity. A third drives the real client-go correlator with production options and worst-case exporter-shaped traffic to confirm the backstop never drops first; it is the intended tripwire for `client-go` bumps. `make test`, `hack/check-generate.sh`, `gofmt -l` and `GOOS=linux CGO_ENABLED=1 go build ./...` are green on this branch.

### DIFF
--- a/cmd/eks-node-monitoring-agent/main.go
+++ b/cmd/eks-node-monitoring-agent/main.go
@@ -152,6 +152,12 @@ func run() error {
 		BaseContext:            func() context.Context { return ctx },
 		Metrics:                server.Options{BindAddress: controllerMetricsAddress},
 		PprofBindAddress:       controllerPprofAddress,
+		// Component-aware event correlation (aggregation keyed on the owning
+		// component; spam filter as a backstop above the exporter's own
+		// budgets). controller-runtime does not shut down a user-provided
+		// broadcaster; this one intentionally lives for the process
+		// lifetime.
+		EventBroadcaster: manager.NewEventBroadcaster(),
 	})
 	if err != nil {
 		logger.Error(err, "failed to create controller manager")
--- a/cmd/eks-node-monitoring-agent/main.go
+++ b/cmd/eks-node-monitoring-agent/main.go
@@ -38,6 +38,7 @@ import (
 	"github.com/aws/eks-node-monitoring-agent/pkg/diagnostic"
 	"github.com/aws/eks-node-monitoring-agent/pkg/manager"
 	"github.com/aws/eks-node-monitoring-agent/pkg/monitor/registry"
+	"github.com/aws/eks-node-monitoring-agent/pkg/reasons"
 
 	// Import monitor packages to trigger auto-registration via init()
 	_ "github.com/aws/eks-node-monitoring-agent/monitors/kernel"
@@ -344,6 +345,20 @@ func run() error {
 		// Initialize monitoring manager
 		logger.Info("initializing monitoring manager")
 		monitorMgr := manager.NewMonitorManager(hostname, nodeExporter)
+
+		// Every monitor name must be a declared component in reasons.yaml: the
+		// monitor name is the fallback event-budget key for reasons that cannot
+		// be resolved by name (parameterized templates render to unregistered
+		// strings), so an undeclared monitor would emit outside the component
+		// capacity ledger and break the derived node-ceiling arithmetic.
+		monitorNames := make([]string, 0, len(enabledMonitors))
+		for _, mon := range enabledMonitors {
+			monitorNames = append(monitorNames, mon.Name())
+		}
+		if err := reasons.ValidateMonitorComponents(monitorNames...); err != nil {
+			logger.Error(err, "monitor component validation failed")
+			return err
+		}
 
 		// Register all monitors with the manager
 		for _, mon := range enabledMonitors {
--- a/pkg/manager/broadcaster_test.go
+++ b/pkg/manager/broadcaster_test.go
@@ -1,0 +1,78 @@
+package manager
+
+import (
+	"sync/atomic"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+// countingSink counts every event write the correlator lets through.
+type countingSink struct {
+	ops atomic.Int64
+}
+
+func (s *countingSink) Create(event *corev1.Event) (*corev1.Event, error) {
+	s.ops.Add(1)
+	return event, nil
+}
+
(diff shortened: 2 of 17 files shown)


cmd/eks-node-monitoring-agent/main.go
```diff
@@ -345,6 +352,20 @@ func run() error {
...
+		// monitor name is the fallback event-budget key for reasons that cannot
+		// be resolved by name (parameterized templates render to unregistered
+		// strings), so an undeclared monitor would emit outside the component
+		// capacity ledger and break the derived node-ceiling arithmetic.
+		monitorNames := make([]string, 0, len(enabledMonitors))
+		for _, mon := range enabledMonitors {
+			monitorNames = append(monitorNames, mon.Name())
+		}
+		if err := reasons.ValidateMonitorComponents(monitorNames...); err != nil {
+			logger.Error(err, "monitor component validation failed")
```
Pavani-Panakanti, line 365: we can add a registry-driven test that iterates every registered monitor and asserts IsComponent(mon.Name()), so issue is caught at build time, not blocker though

pkg/reasons/meta.go
```diff
@@ -4,24 +4,68 @@ package reasons
...
+// arithmetic. Call it at startup and fail loudly on error.
+func ValidateMonitorComponents(names ...string) error {
+	var unknown []string
+	for _, n := range names {
+		if !IsComponent(n) {
+			unknown = append(unknown, n)
+		}
+	}
+	if len(unknown) > 0 {
+		return fmt.Errorf("monitors %v are not declared components in pkg/reasons/reasons.yaml; add Component entries for their reasons (see .kiro/specs/probe-framework-event-limits-spec.md section 3.2)", unknown)
```
Pavani-Panakanti, line 64: remove references to ".kiro/specs/probe-framework-event-limits-spec.md section 3.2" - no such file exists

tools/codegen-reasons/main.go
```diff
@@ -29,11 +49,25 @@ func main() {
...
+	components := make([]string, 0, len(componentSet))
+	for c := range componentSet {
+		components = append(components, c)
+	}
+	sort.Strings(components)
+	if len(components)+len(reservedComponents) > maxComponents {
+		panic(fmt.Errorf(
+			"reasons.yaml declares %d distinct components %v plus %d reserved %v, exceeding the declared platform capacity N=%d; "+
+				"raising N is a deliberate ledger change that recomputes the node-global event ceiling — "+
+				"see .kiro/specs/probe-framework-event-limits-spec.md section 2",
```
Pavani-Panakanti, line 68: same as earlier remove references to ".kiro/specs/probe-framework-event-limits-spec.md section 2"

pkg/manager/manager.go
```diff
@@ -157,17 +158,39 @@ func (m *MonitorManager) exportCondition(ctx context.Context, monitorName string
...
+	if meta, ok := reasons.ByName(reason); ok {
+		return meta.Component()
+	}
+	return monitorName
+}
+
+// SendCondition sends a condition to the exporter based on severity. The
+// component is resolved from the monitor name alone, so prefer the internal
+// path when a reason-specific component may apply.
 func (m *MonitorManager) SendCondition(ctx context.Context, condition monitor.Condition, conditionType corev1.NodeConditionType) error {
```
Pavani-Panakanti, line 180: SendCondition seems to have no caller and dead code as calls to it are moved to sendCondition func below

pkg/manager/broadcaster_test.go
```diff
@@ -0,0 +1,78 @@
...
+// TestBroadcasterBackstopNeverFiresBelowExporterCeiling drives the real
+// client-go correlator, configured exactly as production, with worst-case
+// exporter-shaped traffic: every declared component bursting its full
+// per-type budget with distinct messages. The exporter's shaping bounds this
+// traffic to the node ceiling per event type, and the correlator's spam
+// filter is sized at twice that ceiling, so not one event may be skipped.
+//
+// This test is the designated tripwire for k8s.io/client-go bumps: the
+// backstop-sizing argument depends on correlator internals (default spam key
+// composition, aggregation before filtering) verified at v0.36.2. If a bump
```
Pavani-Panakanti, line 41: go.mod is at k8s.io/client-go v0.36.3 - can we validate it on v0.36.3 if any changes around this. If not then we can update comment

Pavani-Panakanti:
Few other points:

* We send an event every poll cycle while a check is failing. At 30s that's ~120/hr, but the limit is 25 then ~12/hr, so ~108/hr just get dropped. Events last 1 hour and re-sending only refreshes the same one, so we don't need to send every cycle. Can we evaluate refreshing once every few minutes or some longer cadence instead of every poll?
* Do we need the separate "resolved" event? The node condition already goes back to healthy on its own, and the old failure event disappears after an hour once we stop sending it. Want to see if we should skip it and save some budget or anything else I am missing 
* Can we add a short doc on how to add a new component to the repo itself ? The limit (25 then ~12/hr per component) is only in code. It would be helpful to add some guidelines - use a Fatal condition for ongoing problems (no event spam), and check nma_events_dropped_total to make sure nothing's being dropped. Some readme or doc would help

Issue #243, opened 2026-08-24
bug(dcgm): agent cannot reach dcgm-server on clusters without hostPort support (no hostNetwork option)


robsoned:
**What happened**:

The node agent connects to the dcgm-server host engine at `localhost:5555`
and relies on the dcgm-server DaemonSet's `hostPort: 5555` to translate
that to the pod. On clusters whose CNI does not implement hostPort (no
portmap plugin in the CNI chain), that DNAT never exists, so the agent can
never connect. Every reconcile fails with:

    failed to initialize DCGM: error connecting to nv-hostengine:
    Host engine connection invalid/disconnected

and the node latches AcceleratedHardwareReady=False with reason DCGMError,
even though the host engine is healthy: the same dcgmi query succeeds
against the dcgm-server pod IP, and from inside the dcgm-server pod.

Verified from the agent pod (hostNetwork) with the bundled dcgmi:

    localhost:5555   -> connect fails
    127.0.0.1:5555   -> connect fails
    <node-ip>:5555   -> connect fails
    <pod-ip>:5555    -> connects, reports all GPUs

The chart offers no way to avoid the hostPort path: dcgmAgent has no
hostNetwork value, and the managed addon schema (additionalProperties:
false) rejects unknown keys, so cluster operators cannot fix this through
configuration.

**What you expected to happen**:

The chart exposes `dcgmAgent.hostNetwork` so the host engine can bind
port 5555 on the node directly, removing the hostPort/NAT dependency.
NVIDIA's own DCGM components (e.g. the GPU Operator's dcgm-exporter)
run hostNetwork for the same reason.

**How to reproduce it (as minimally and precisely as possible)**:

1. EKS cluster with GPU nodes and a CNI configuration without the
   portmap plugin (hostPort not implemented).
2. Install the eks-node-monitoring-agent addon v1.7.0-eksbuild.1.
3. The agent logs the connection error above on every reconcile and sets
   AcceleratedHardwareReady=False (reason DCGMError) on all GPU nodes.
4. Patch the dcgm-server DaemonSet with `hostNetwork: true` and
   `dnsPolicy: ClusterFirstWithHostNet`: the agent connects on its next
   reconcile and the condition recovers.

Related: #205 cover pointing the agent at an external hostengine
(`nodeAgent.dcgmAddress`), which avoids hostPort only when the bundled
server is disabled and an external hostengine exists. This issue covers
the bundled dcgm-server, which has no Service and churning pod IPs, so
[...]

PR #244, opened 2026-08-24
fix(chart): support hostNetwork for the dcgm-server pod

robsoned:
Issue #243

**Description of changes**:

Adds an opt-in `dcgmAgent.hostNetwork` value (default `false`) to the dcgm-server DaemonSet template, values, and addon schema. When set, the dcgm-server pod runs with `hostNetwork: true` and `dnsPolicy: ClusterFirstWithHostNet`, letting the host engine bind port 5555 on the node directly.

The agent connects to the host engine at `localhost:5555`, relying on the dcgm-server DaemonSet's `hostPort: 5555` to NAT that to the pod. On clusters whose CNI has no portmap plugin, that NAT never exists, so every reconcile fails (`Host engine connection invalid/disconnected`) and the node latches `AcceleratedHardwareReady=False` (reason `DCGMError`) even when the hardware is healthy. `hostNetwork` removes the hostPort dependency entirely — the same approach NVIDIA's own dcgm-exporter (GPU Operator) uses.

**Testing Done**:

- Live validatio from the agent pod (hostNetwork), `localhost:5555`, `127.0.0.1:5555`, and `<node-ip>:5555` all failed to connect; only the dcgm-server pod IP succeeded.
- Patched the dcgm-server DaemonSet with `hostNetwork: true` / `dnsPolicy: ClusterFirstWithHostNet`; agent connected on the next reconcile and the node condition recovered.
- Re-validated with the chart change (`dcgmAgent.hostNetwork: true`) end to end — same recovery, no manual patch required.
- `helm lint` and `helm template` pass with the value present and absent (default off, no behavior change for existing installs).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license

### DIFF
--- a/charts/configuration.schema.json
+++ b/charts/configuration.schema.json
@@ -129,6 +129,11 @@
                     "description": "Name of an existing PriorityClass assigned to the dcgm-exporter pod.",
                     "default": "system-node-critical"
                 },
+                "hostNetwork": {
+                    "type": "boolean",
+                    "description": "Run the dcgm-server pod on the host network so nv-hostengine binds node port 5555 directly. Required on clusters whose CNI does not implement hostPort.",
+                    "default": false
+                },
                 "resources": {
                     "$ref": "#/definitions/Resources"
                 },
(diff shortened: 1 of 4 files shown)


erezzarum:
A safer approach will be to create a `Service` with `internalTrafficPolicy` set to `Local` and use that as the dcgm agent address the NMA agent uses.

Issue #224, opened 2026-08-13
bug(networking-monitor): On EKS Automode nodes with 2 vCPU and with high pod churn, NMA network monitor sweep briefly consumes both the CPU which is introducing latency into the system


ozashai:
**What happened**:
On EKS Automode, for a worker node with 2 vCPU and high pod churn, NMA network monitor sweep is briefly consuming both the vCPUs which is starving the network processing (softirq) for ~0.5–1s. Any new outbound connection (DNS / TCP / TLS) that lands in that window sees a latency spike, then recovers. 

**What you expected to happen**:
The NMA network monitor sweep should not introduce latency into the system.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create or use an existing EKS automode cluster.
2. Create a nodeclass and nodepool and use instances with 2vCPU (e.g., c6g.large)
3. Create a latency probe deployment and load deployment that introduces pod churn into the system. The probe should be programmed to open a new HTTPS connection to an endpoint (say sts.amazonaws.com) per second and should record DNS / TCP-connect / TLS phase times. The load introduced to the worker node should maintain CPU utilization at 60-75%. 
For e.g.,
containers:
        - name: probe
          image: public.ecr.aws/docker/library/alpine:3.20
          resources: { requests: { cpu: "10m", memory: "32Mi" } } 
          command: ["/bin/sh","-c"]
          args:
            - |
              apk add --no-cache curl coreutils >/dev/null 2>&1
              TARGET="${TARGET:-sts.us-east-1.amazonaws.com}"
              echo "ts_utc,epoch,dns_s,connect_s,appconnect_s,total_s,http_code"
              while true; do
                TS=$(date -u +%Y-%m-%dT%H:%M:%S); EP=$(date +%s.%N)
                OUT=$(curl -4 -s -o /dev/null --connect-timeout 5 --max-time 10 \
                      -w '%{time_namelookup},%{time_connect},%{time_appconnect},%{time_total},%{http_code}' \
                      "https://${TARGET}/" 2>/dev/null || echo "ERR,ERR,ERR,ERR,000")
                echo "${TS}Z,${EP},${OUT}"; sleep 1
              done
4. NMA sweeps happen at 5 minute intervals. Let the test run for approximately 15 minutes.

**Environment**:
- AWS Region: us-east-1
- Instance Type(s): c6g.large
- Cluster Kubernetes version: 1.35

[...]

PR #257, opened 2026-09-01
fix(diagnostic): rate-limit console writes to avoid serial-console stalls

wendyz-am:
**Issue #, if available**: Relates to #224

**Description of changes**:
This change wraps the console writer in a rate limiter (golang.org/x/time/rate, already a dependency) that feeds bytes to the device at 32 KB/s in ~2KB chunks — below the drain rate. This bounds every contiguous write regardless of section size, so no single write occupies the serial line long enough to stall the network path. RunOnce and the emitted content are unchanged; only the write rate is paced. Pacing is always on; the tradeoff is a slightly longer total cycle (~2.1s vs ~1.7s), which stays well under the 5-minute log interval.

**Testing Done**:
- Unit tests (pkg/diagnostic/logger_test.go): writes above the chunk size are split into correctly sized chunks with bytes arriving intact and in order, sub-chunk writes pass through as a single unsplit write, the burst-equals-chunk invariant that makes the spacing work holds, throughput is throttled, the underlying write error surfaces unwrapped, and NewDiagnosticLogger actually wraps its writer. 
- Root-cause isolation (non-Auto 2-vCPU, arm64, under load): a controlled back-to-back write of a full cycle's bytes to /dev/console reproduced DNS/TCP latency spikes (up to ~600–900ms) coinciding with the write; running the collectors without the write produced none; splitting the same bytes into rate-limited chunks eliminated the spikes.
- Symptom reproduction (Auto Mode 2-vCPU, arm64, #224 setup): latency spikes reproduce on the 5-minute NMA cycle (up to ~1.1s total), confirming the issue on the reported configuration.
- On-node smoke test: built and ran the patched agent on a 2-vCPU node; all 10 sections are still emitted, now fed to the console at the paced rate.

### DIFF
--- a/pkg/diagnostic/logger.go
+++ b/pkg/diagnostic/logger.go
@@ -14,6 +14,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"golang.org/x/time/rate"
 	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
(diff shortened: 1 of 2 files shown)


pkg/diagnostic/logger.go
```diff
@@ -73,12 +74,82 @@ const (
 	// unit journal is buffered into memory each cycle and grows unbounded
 	// over the node's lifetime. Chosen to comfortably exceed the section budget.
 	journalMaxLines = "2000"
+
+	// consoleBytesPerSecond caps the rate at which diagnostic output is handed
```
prasad0896: This is too descriptive, verbose comment and does not add much value to the code. Ideally we only want comments explaining the "why" not the how.

wendyz-am: That makes sense, just shorten this comment and some of the other ones in 5fac3f0.

prasad0896 (approved):

PR #258, opened 2026-09-01
allow disabeling reasons

grosser:
**Description of changes**:
we want to have an emergency off button if a reason has false-positives,
for example as we saw with IPAMDNotReady [here](https://github.com/aws/eks-node-monitoring-agent/pull/216)

**Testing Done**:

### DIFF
--- a/pkg/config/monitor.go
+++ b/pkg/config/monitor.go
@@ -19,6 +19,10 @@ type MonitorSettings struct {
 	Enabled                      *bool    `yaml:"enabled,omitempty" json:"enabled,omitempty"`
 	AllowedIPTablesChains        []string `yaml:"allowedIPTablesChains,omitempty" json:"allowedIPTablesChains,omitempty"`
 	ExcludedInterfaceNameRegexps []string `yaml:"excludedInterfaceNameRegexps,omitempty" json:"excludedInterfaceNameRegexps,omitempty"`
+	// DisabledReasons lists reason names (e.g. "IPAMDNotReady") whose
+	// conditions should be suppressed. Parameterized reasons are specified in
+	// their rendered form (e.g. "NvidiaXID79Error").
+	DisabledReasons []string `yaml:"disabledReasons,omitempty" json:"disabledReasons,omitempty"`
 }
 
 // IsEnabled returns true if the monitor is enabled.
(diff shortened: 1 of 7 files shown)


pkg/manager/manager.go
```diff
@@ -124,6 +137,12 @@ func (m *MonitorManager) runLoop(ctx context.Context) error {
 func (m *MonitorManager) exportCondition(ctx context.Context, monitorName string, condition monitor.Condition) error {
 	logger := log.FromContext(ctx).WithValues("source", monitorName, "condition", condition)
 
+	// Drop conditions whose reason has been disabled by configuration.
+	if _, ok := m.disabledReasons[condition.Reason]; ok {
```
wendyz-am, line 141: Was dropping the metric deliberate? This sits above `conditionCount.WithLabelValues(...).Add(1)`, so a disabled reason also stops being counted in `problem_condition_count`.

For an emergency mute, I think we might want the opposite: stop exporting, keep counting. While a reason is muted, the counter is what tells us whether the false positive is still firing, and whether an upstream fix has landed so we can un-mute.

pkg/config/monitor.go
```diff
@@ -100,6 +104,20 @@ func (mc *MonitorConfig) GetExcludedInterfaceNameRegexps() []string {
...
 
+// GetDisabledReasons returns the aggregated list of reasons disabled across
+// all monitors. Reason names are globally unique across monitors, so an entry
+// applies wherever that reason is emitted.
+func (mc *MonitorConfig) GetDisabledReasons() []string {
+	if mc == nil || mc.Monitors == nil {
+		return nil
+	}
+	var disabled []string
+	for _, settings := range mc.Monitors {
```
wendyz-am, line 115: The `_` drops the monitor name here, and the manager filters on reason alone, so disabledReasons nested under kernel-monitor will still silence a networking reason.

I think keeping the per-monitor shape is right for future-proofing (reason names are unique today, but if that ever changes we can tighten the filter without a config break). Could we just note in the README or doc comment that the filter is global for now and the nesting is reserved?

wendyz-am (commented):
Thank you, this is a useful feature. Left two inline notes.

Also this needs a rebase — #264 landed on main and changed Validate / LoadMonitorConfig to take a knownPluginNames param, and deleted the KnownPluginNames var.

pkg/manager/manager.go
grosser: I wanted our monitoring to also "shut up about it", but I can use regular muting for that, so should be fine to keep it here.

pkg/config/monitor.go
grosser: good catch, fixed ... rather have it working exactly right then document weird edge-cases :)

grosser:
all fixed 💇

PR #270, opened 2026-09-14
feat(nvidia): configure DCGM power policy threshold

dwo:
**Issue #, if available**:

N/A

**Description of changes**:

NMA currently registers the DCGM power policy without supplying a threshold, so [go-dcgm uses its generic 250 W default](https://github.com/NVIDIA/go-dcgm/blob/v1.4601.1/pkg/dcgm/policy.go#L55-L65).
Normal board power on newer accelerators can exceed that threshold; for example, [AWS documents a 1000 W power cap for P6-B200](https://docs.aws.amazon.com/dlami/latest/devguide/p6-support-dlami.html).
This can produce `NvidiaPowerError` warnings during normal workloads.

This change adds the optional `nodeAgent.monitors.nvidia.dcgmPowerThresholdWatts` setting to the Helm chart and EKS add-on configuration schema.
When configured, NMA preserves the current DCGM policy action, validation, and non-power thresholds while replacing the power threshold before registering its listener.
When omitted, the existing policy setup remains unchanged.

```yaml
nodeAgent:
  monitors:
    nvidia:
      dcgmPowerThresholdWatts: 1000
```

**Testing Done**:

- `go test ./pkg/config/...`
- `helm lint charts/eks-node-monitoring-agent --set nodeAgent.monitors.nvidia.dcgmPowerThresholdWatts=1000`
- Rendered the monitor ConfigMap with the configured value.


wendyz-am:
Thank you for this PR. One note though, `uint32` assertions are on the thermal / retired-pages values `GetPolicyForGroup` hands back as `interface{}`. They pass today, but that Go type is go-dcgm's private choice and we bump the dependency automatically. If it changed, the error would propagate out of Reconcile and permanently kill XID / ECC / NVLink / PCIe detection on the node. Upstream's policyThresholdUint32 falls back to the default value instead of erroring, could we mirror that (log + default, never fail Reconcile)?

dwo:
@wendyz-am Took a stab at mirroring that: how does it look?

wendyz-am:
looks good to me, thank you

wendyz-am (approved):

PR #273, opened 2026-09-15
fix(nvidia): detect GPU Operator device path for device count

pjanisze98:
**Issue #, if available**:
N/A

**Description of changes**:
This fix is coming as a follow up to the recently released feature that lets us use an external DCGM server and thus have NVIDIA GPU operator compatibility. There is a now a bug with the GPU Count check on nodes managed by the operator.

When the NVIDIA GPU Operator manages the driver as a container, nvidia device files are created under the operator's `hostPaths.driverInstallDir` (default `/run/nvidia/driver`) instead of the standard `/dev/`. The existing `GetNvidiaFSDeviceCount()` only checked `/dev/nvidia[0-9]*`, causing a false `AcceleratedHardwareReady=False` condition on GPU Operator managed nodes. Example: "DCGM detected 8 GPUs but 0 nvidia device files were detected"

This PR adds a `nvidiaDevDir()` helper that resolves the correct device directory:
1. Explicit `NVIDIA_DRIVER_ROOT` env var override, for clusters where [`hostPaths.driverInstallDir`](https://github.com/NVIDIA/gpu-operator/blob/main/deployments/gpu-operator/values.yaml#L35) has been changed from the default
2. Auto detect `/run/nvidia/driver/dev/` if the directory exists since its the GPU operator default
3. Fall back to `/dev/` (traditional host installed driver)

Also exposes `nvidiaDriverRoot` as an optional Helm value for the override case.

**Testing Done**:

- Added `TestGetNvidiaFSDeviceCount` with 4 cases: standard path, GPU Operator auto detect, explicit override, no devices
- Added `TestNvidiaDriverRoot` for the env var helper
- All existing tests pass (verified via `go test`)
- Helm lint passes
- Verified on a node running GPU Operator that `/run/nvidia/driver/dev/` contains `nvidia0` through `nvidia7` for a node expected to have 8 GPUs

Authored-By: Paul Janiszewski <pjaniszewski@pinterest.com>
```

### CODE REVIEW

**Review by wwvela — COMMENTED** [review]
Thank you for this fix. Changes look good to me. We will merge once the tests pass

**Review by wwvela — DISMISSED** [review]

**Comment by wwvela** [comment]
Could you run `make generate` to fix the outdated README.md to unblock the check-generate ci failure?

**Comment by prasad0896** [comment]
Good find -- the operator's containerized driver landing devices outside /dev is exactly the kind of gap the external-DCGM path opened up.

Before the path logic matters, the mount has to reach us. The GPU Operator's driver container mounts /run/nvidia with mountPropagation: Bidirectional and creates /run/nvidia/driver as a mount from inside it, while our host-root volumeMount sets no mountPropagation and so defaults to None. A container that started before that mount existed never sees it, and the agent is system-node-critical so it schedules early on a new node. The operator's own driver daemonset sets HostToContainer on the identical /host mount. Can you confirm the nvidia0..7 check was made from inside the agent container rather than on the host, and add:

```yaml
    - name: host-root
      mountPath: /host
      mountPropagation: HostToContainer
```

Second, nvidiaDriverRoot needs an entry in charts/configuration.schema.json. NodeAgent there is additionalProperties: false, so the managed add-on rejects the key outright instead of ignoring it, which puts the override out of reach for anyone not installing the chart directly. dcgmAddress and dnsPolicy went in when that feature landed -- same shape:

```json
    "nvidiaDriverRoot": {
        "type": "string",
        "description": "Override for the NVIDIA driver root path. Only needed when the GPU Operator uses a non-default driver install directory.",
        "default": ""
    },
```

Smaller one: the operator directory is selected on os.Stat alone, so a /run/nvidia/driver/dev that exists but is empty shadows /dev and produces the same 0-device message this is fixing. Globbing both and taking the one with devices would close that.

prasad0896 (changes requested):
added comment above

pjanisze98:
Comment addressed!

prasad0896:
Thanks -- all three are in, and returning the resolved directory so the mismatch message names the path it actually searched is better than what I asked for.

check-generate check is red: e2e/setup/manifests/agent.tpl.yaml is rendered from the chart by
make update-e2e-manifests, so the daemonset mountPropagation change left it stale. A make generate and a commit should clear it and make the PR ready to merge.

